PR #1, opened 2022-05-07
chore: setup checkpoint code

perfectmak:
This basically moves the checkpoint code from sx-api into this repository and setups the package.json files to allow proper importing of the library and it's type definition.

This is mostly for tracking purposes, because reviewing it is basically re-reading most of the code from sx-api.

This also resolves https://github.com/snapshot-labs/sx-api/issues/55


bonustrack (approved):
Looks good

PR #17, opened 2022-05-13
ci: release npm package on main branch

perfectmak:
This setups a github action to trigger releasing a package when merged
to master and the version in the package.json file changes.


.github/workflows/build-test-release.yml
```diff
@@ -0,0 +1,27 @@
...
+          cache: 'yarn'
+      - run: yarn install
+      - run: yarn build
+      - run: yarn test
+      - id: publish
+        name: Release to NPM
+        if: github.ref == 'refs/heads/master'
+        uses: JS-DevTools/npm-publish@v1
+        with:
+          token: ${{ secrets.NPM_TOKEN }}
```
perfectmak, line 20: A valid NPM_TOKEN secret is required to be setup in this repository for this to work.

bonustrack: Would this publish on npm using beta version too?

perfectmak: It should use the version in the package.json file as the `latest` release version when publishing. Or do you mean something else?

perfectmak: I've included the beta tag now.

PR #35, opened 2022-06-05
feat: generate sample query in graphiql

### LINKED ISSUE #24: Generate GraphQL example query
It would be great to have an example query when user start a checkpoint instance in the GraphiQL console, we could use the first entity in the GraphQL schema to generate an example, for SX API it would look like this:
```gql
query {
  spaces(first: 10) {
    id
    name
    about
    voting_delay
    voting_period
  }
}
```
Instead of showing "Welcome to GraphiQL..." default message here: https://goerli-api-1.snapshotx.xyz/

perfectmak:
This generats an example query based on the first defined entity in the
schema argument to Checkpoint.

Resolves #24 .

> Example screenshot from checkpoint-template.

Screenshot 2022-06-06 at 12 04 48 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "express-graphql": "^0.12.0",
     "graphql": "^16.5.0",
     "graphql-fields": "^2.0.3",
+    "json-to-graphql-query": "^2.2.4",
     "mysql": "^2.18.1",
     "pino": "^7.11.0",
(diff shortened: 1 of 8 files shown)


bonustrack (approved):
Looks great! AFAIK we actually not support nested objects/entity yet but that's good to have it working already for the example query

PR #36, opened 2022-06-06
feat: configure starknet provider url

### LINKED ISSUE #8: Add option in the checkpoint config to change provider url
Currently Checkpoint works only with default nodes, we should add an option to make it possible to change the node, this will allow us to use things like https://github.com/snapshot-labs/starknet-proxy to sync block faster

perfectmak:
This updates the checkpoint configuration options to include a base url
as opposed to just specifying the network name. This will allow more
flexibility and options for StarkNet data providers.

Resolves #8


bonustrack (approved):
Looks good!

PR #37, opened 2022-06-06
fix: make sortBy fields a graphql enum

### LINKED ISSUE #9: Enforce options for sortBy selector
I've noticed it's possible to do SQL injection example:
https://goerli-api-1.snapshotx.xyz/?query=query%20%7B%0A%20%20_metadatas%20%7B%0A%20%20%20%20id%0A%20%20%20%20value%0A%20%20%7D%0A%20%20votes%20(first%3A%202%2C%20orderBy%3A%20%22id%3B%20SELECT%20*%20FROM%20votes%20ORDER%20BY%20id%22%2C%20orderDirection%3A%20asc)%20%7B%0A%20%20%20%20voter%0A%20%20%20%20space%0A%20%20%20%20choice%0A%20%20%20%20created%0A%20%20%7D%0A%7D%0A
We should change "orderBy" to accept only defined fields like this: 

```gql
 query {
   votes (orderBy: created)
 }
```
 instead of this
 ```gql
 query {
   votes (orderBy: "created")
 }
 ```

perfectmak:
This makes the sortBy fields for multi record queries to be generated as
enums. This will also help fix a potential sql injection issues that
allows users to insert arbitrary orderBy strings in queries.

Resolves #9 

Screenshot 2022-06-06 at 9 43 37 pm

### DIFF
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -17,6 +17,7 @@ import {
   GraphQLScalarType,
   GraphQLSchema,
   GraphQLString,
+  isLeafType,
   Source
 } from 'graphql';
 import { AsyncMySqlPool } from '../mysql';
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
Looks great!

PR #66, opened 2022-06-23
fix: enable synchronous logging

### LINKED ISSUE #60: Enable synchronous logging for Pino
Currently Pino don't return log when they come, we should change that.

perfectmak:
This enables synchronous logging for pino, which will ensure that log
outputs are not buffered but output instantly. This will help reduce
confusion when the writer functions are using the default console logger
or some other synchronous logger of their choice.

Resolves #60 

Also includes fix: enable pretty logging without dependency
This enables pretty logging without explicitly requiring the users of
the library to install pino-pretty.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "json-to-graphql-query": "^2.2.4",
     "mysql": "^2.18.1",
     "pino": "^8.0.0",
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
Looks good!

PR #112, opened 2022-08-13
choosing the node one wants to use for indexing rather than use default ones only

envusr:
this version of checkpoint allows to choose the node one wants to use for indexing starknet data rather than having to use default ones
for this network_base_url in CheckpointConfig has been replaced by provider_config which is a simple starknet ProviderOptions object ({ baseUrl: URL } for a sequencer node, { rpc: { nodeUrl: URL }} for an rpc node, etc)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "mysql": "^2.18.1",
     "pino": "^8.3.1",
     "pino-pretty": "^8.1.0",
-    "starknet": "^3.7.0"
+    "starknet": "^4.3.0"
   },
   "devDependencies": {
     "@types/bn.js": "^5.1.0",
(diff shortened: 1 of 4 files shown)


bonustrack:
@envusr Thanks for the PR, did you tried to sync a contract with it? I dont think it would work out of the box cuz the API request and response format changed. Also types should but updated

envusr:
hey @bonustrack yes I tried, and it worked, maybe some features are not working anymore, a double check is the best way to know
but contracts new events fetching works perfectly
and I have updated the types

envusr:
it actually doesn't work
checkpoint is launched but can't find any event

Sekhmet:
PR that adds support for RPC: https://github.com/snapshot-labs/checkpoint/pull/154
(and drops Sequencer as-for-now, because in newer version of starknet sequencer would require making additional requests).

Sekhmet:
As of `0.1.0-beta.13` it's configurable and RPC is supported (and Sequencer support was dropped): https://github.com/snapshot-labs/checkpoint/commit/b5bd70feddc6b2a29168000ef5ef164a4852d252

PR #145, opened 2022-09-24
feat: add support for lists

Sekhmet:
Adds support for GraphQL list. Those are represented using `JSON` column type. Those are not indexed and can't be used in `where` for now.

## Test plan

- Test using sx-api PR: https://github.com/snapshot-labs/sx-api/pull/150


bonustrack (approved):
tACK

PR #150, opened 2022-09-30
feat: unwrap nonNull

Sekhmet:
If you make nested entity non nullable it will be wrapped in GraphQLNonNull so we need to unwrap it when using it.

## Test plan

- Change `author: User` to `author: User!` in sx-api's Proposal schema.
- Run query below.

```gql
{
  proposals(where: {space: "0x00847ae39833c61fe964bfac857ad8c5fa261ec4c716c61b8f28c4ae61fe376b"}) {
    id
    title
    author {
      id
      vote_count
    }
  }
}
```


bonustrack (approved):
tACK

PR #153, opened 2022-11-10
feat: add support for BigInt and decimal types

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/checkpoint/issues/152

This PR adds support for `BigInt` and decimal types (`Decimal` and `BigDecimal`).
Current mappings are:
- `BigInt` -> `BIGINT`
- `Decimal` -> `DECIMAL(10, 2)`
- `BigDecimal` -> `DECIMAL(20, 8)`.

Decimal types are configurable so you can provide different precision/decimals values (or add more variants):
```js
const checkpoint = new Checkpoint(config, writer, schema, {
  decimalTypes: {
    Decimal: {
      p: 14,
      d: 10
    },
    BigDecimal: {
      p: 20,
      d: 8
    },
    EvenBiggerDecimal: {
      p: 40,
      d: 16
    }
  }
});
```

## Usage

In your schema declare types that you use (either default ones, or the ones you added) and use those in your types as needed
```gql
scalar BigInt
scalar Decimal
scalar BigDecimal

type User {
  votes_count: BigInt
}  
```

When writing data pass your data in format `mysql` package supports (Number/String/BigInt), making sure you are not losing precision on-write.

GraphQL will return those values as string, you can query those values with where filter as with `Int` type (gt/gte/lt/lte etc. are supported), but you should use strings there to make sure precision is not lost.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -41,7 +41,7 @@ export default class Checkpoint {
     this.config = config;
     this.writer = writer;
     this.schema = schema;
-    this.entityController = new GqlEntityController(schema);
+    this.entityController = new GqlEntityController(schema, opts);
 
     const providerConfig = this.config.network_base_url
       ? { baseUrl: this.config.network_base_url }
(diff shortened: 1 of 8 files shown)


bonustrack (approved):
utACK

PR #154, opened 2022-11-11
feat: use RPC instead of Sequencer

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/checkpoint/issues/54

This PR **replaces** Sequencer with RPC for indexing.

**Limitations:**
 - In latest versions of `starknet` sequencer provider doesn't return full transactions in blocks response and there is no way to opt-out of it as-is. This means we would need to add extra request to fetch every transaction in block. As of now this PR drops support for Sequencer and will work with RPC only.
 - ~RPC doesn't return receipts in the block response, this means we have to query receipts manually which adds extra requests.~ solved by using `getEvents` call.
 - because we use `getBlockWithTxs` + `getEvents` calls we don't have access to receipts in the responses (unless we add extra request per transaction for it) so writers no longer receive `receipt`.
 
 ## Test plan
 - Run `sx-api` with `network_node_url` in config.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -1,4 +1,4 @@
-import { api, RpcProvider } from 'starknet';
+import { RpcProvider } from 'starknet';
 import { starknetKeccak } from 'starknet/utils/hash';
 import { validateAndParseAddress } from 'starknet/utils/address';
 import Promise from 'bluebird';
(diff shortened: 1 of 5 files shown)


src/checkpoint.ts
```diff
@@ -242,13 +236,21 @@ export default class Checkpoint {
...
-    // @ts-ignore
-    for (const receipt of block.transaction_receipts || []) {
-      await this.handleTx(block, block.transactions[receipt.transaction_index], receipt);
+    const receipts = await Promise.all(
+      block.transactions.map(tx => {
+        if (!tx.transaction_hash) return null;
+
+        return this.provider.getTransactionReceipt(tx.transaction_hash);
+      })
+    );
```
Sekhmet, line 250: Additional requests required when using RPC.

bonustrack: This might result in a big slowdown to sync, isn't there a way to get all events from a block within a single request?

Sekhmet: Maybe with `starknet_getEvents` 🤔

bonustrack: @Sekhmet Yes
image

Sekhmet: That works, but we now miss receipt as those can only be fetched by those one-per-tx requests in RPC client so writers no longer receive them. Is that okay?

bonustrack: @Sekhmet Yes I guess we can drop this feature for now. Does the event return the tx hash?

Sekhmet: TX hash is in both events response and blocks response.

bonustrack (approved):
tACK

PR #156, opened 2022-11-12
feat: add templates support

Sekhmet:
## Summary

This PR allows consumers to modify sources on the fly, when other events arrive.
Writers now receive `instance` argument, which is Checkpoint instance itself. In this case developers can use `executeTemplate` to use predefined template from config.

Closes: https://github.com/snapshot-labs/checkpoint/issues/52

## Test plan

Update sx-api config to:
```json
{
  "network_node_url": "RPC",
  "sources": [
    {
      "contract": "0x2121c175922cef8870b6b6b956d01a7ac75af034dfbb9135698f88644b17ea3",
      "start": 345862,
      "events": [
        {
          "name": "space_deployed",
          "fn": "handleSpaceCreated"
        }
      ]
    }
  ],
  "templates": {
    "Space": {
      "events": [
        {
          "name": "proposal_created",
          "fn": "handlePropose"
        },
        {
          "name": "vote_created",
          "fn": "handleVote"
        }
      ]
    }
  }
}
```

Call this function in `handleSpaceCreated`:
```js
instance.executeTemplate('Space', {
  contract: item.id,
  start: block.block_number
});
```

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -8,6 +8,7 @@ import { GqlEntityController } from './graphql/controller';
 import { createLogger, Logger, LogLevel } from './utils/logger';
 import { AsyncMySqlPool, createMySqlPool } from './mysql';
 import {
+  ContractSourceConfig,
   Block,
   FullBlock,
   Transaction,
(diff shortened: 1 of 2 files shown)


bonustrack:
Haven't been able to reproduce but I think there is a risk that some block get omitted when we query next checkpoints with the function `getNextCheckpointBlocks` and a new contract get added with `addSource` and this contract includes txs within the latest range of checkpoints we got. 

Also it would be great if we can add source from the config.json file directly, this is how TheGraph is doing, they call it templates https://github.com/snapshot-labs/snapshot-subgraph/blob/master/subgraph.optimism.yaml#L53-L71

Sekhmet:
Thanks, updated `addSource` to invalidate checkpoints cache.

I also exposed templates config, similar to [Subgraph's](https://thegraph.com/docs/en/developing/creating-a-subgraph/#data-source-templates). What do you think of this API? (check testplan for usage).

bonustrack (approved):
tACK

PR #157, opened 2022-11-12
feat: use dataloader to batch single entity queries

Sekhmet:
## Summary

This PR adds [`dataloader`](https://github.com/graphql/dataloader) support for single queries that improves performance of nested queries. Now fetching 10 proposals with spaces nested in the query takes 2 requests instead of 20.

Closes: https://github.com/snapshot-labs/checkpoint/issues/147

## Test plan

Add console.log above `const results = await context.mysql.queryAsync(query, [ids]);`.

Run query:
```gql
{
  proposals(first: 100) {
    id
    proposal_id
    space {
      id
      name
    }
  }
}
```

Only single log should be emitted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "@graphql-tools/schema": "^8.5.1",
     "bluebird": "^3.7.2",
     "connection-string": "^4.3.5",
+    "dataloader": "^2.1.0",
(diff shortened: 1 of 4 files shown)


bonustrack (approved):
tACK

PR #159, opened 2022-11-18
feat: add rpc 0.2.1 compatibility

Sekhmet:
## Summary

RPC 0.2.1 changes pagination format so we need to update the code to be compatible with latest RPC servers.


bonustrack (approved):
utACK

PR #161, opened 2022-11-18
chore: narrow down writer's block param

Sekhmet:
## Summary

We check that block has to be full block before calling writers so we can narrow it down to avoid awkward handling at writer side.


bonustrack (approved):
utACK

PR #162, opened 2022-11-22
feat: expose schema and createGetLoader

Sekhmet:
## Summary

This allows to use checkpoint via ApolloServer (and possibly other clients).

## Test plan

Build checkpoint from this PR and copy files over (instead of linking, as GraphQL depends on instance equality which doesn't work when linked) to sx-api, setup ApolloServer there like this:
```js
import Checkpoint, { LogLevel, createGetLoader } from '@snapshot-labs/checkpoint';

import { ApolloServer } from '@apollo/server';
import { startStandaloneServer } from '@apollo/server/standalone';

// Same checkpoint initialization here as before.

async function run() {
  const server = new ApolloServer({
    schema: checkpoint.getSchema()
  });

  const { url } = await startStandaloneServer(server, {
    listen: { port: 3000 },
    context: async () => {
      const baseContext = checkpoint.getBaseContext();
      return {
        ...baseContext,
        getLoader: createGetLoader(baseContext)
      };
    }
  });

  console.log(`🚀  Server ready at: ${url}`);
}

run();
```

Run sx-api, execute sample query:
```gql
{
  proposals(first: 100) {
    id
    proposal_id
    space {
      id
      name
    }
  }
}
```

## Screenshots
image

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -72,12 +72,14 @@ export default class Checkpoint {
     this.mysqlConnection = opts?.dbConnection;
   }
 
-  /**
-   * Returns an express handler that exposes a GraphQL API to query entities defined
-   * in the schema.
-   *
-   */
-  public get graphql() {
+  public getBaseContext() {
+    return {
+      log: this.log.child({ component: 'resolver' }),
+      mysql: this.mysql
+    };
+  }
+
+  public getSchema() {
     const entityQueryFields = this.entityController.generateQueryFields();
     const coreQueryFields = this.entityController.generateQueryFields([
       MetadataGraphQLObject,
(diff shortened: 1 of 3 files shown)


bonustrack:
Nice, Apollo explore UI looks much better than GraphiQL. 

Not sure what I'm missing in my setup but when I try the test plan I got this error:

> Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of other
relied on modules, use "resolutions" to ensure only one version is installed.

Using resolutions doesnt seem to help, I keep having same error.

Sekhmet:
@bonustrack You have to copy the package manually, linking doesn't work (more context in the test plan).

bonustrack:
Yes I did that, also tried with link but doesnt work for me

Sekhmet:
> Yes I did that, also tried with link but doesnt work for me

Most likely because there is `graphl` in node_modules of checkpoint package after copying. Can you remove it from `node_modules/@snapshot-labs/checkpoint/node_modules`? Alternatively remove it, run yarn and then copy just `dist` directory instead of entire thing.

bonustrack (approved):
@Sekhmet Thanks works well now, tACK

PR #163, opened 2022-11-23
chore: bump starknet to 4.13.1 to fix Infura compatibility

Sekhmet:
4.13.1 no longer hardcodes `/rpc/v0.2` as part of URL so it's compatible with Infura.


bonustrack (approved):
utACK

PR #167, opened 2022-11-24
chore: use shared eslint and prettier configs

Sekhmet:
CI should pass.


package.json
```diff
@@ -11,13 +11,15 @@
...
+    "extends": "@snapshot-labs"
+  },
+  "prettier": "@snapshot-labs/prettier-config",
   "dependencies": {
     "@graphql-tools/schema": "^8.5.1",
     "bluebird": "^3.7.2",
     "connection-string": "^4.3.5",
     "dataloader": "^2.1.0",
-    "dotenv": "^16.0.1",
-    "eslint": "^8.21.0",
```
Sekhmet, line 20: For some reason eslint was a dependency.For some reason those two were deps. Moved them to devDeps.

bonustrack: Sound good, I didn't knew `dotenv` should be a dev dependency, for `eslint` it was probably not here on purpose.

bonustrack (approved):
utACK

PR #171, opened 2022-12-01
feat: use default connections limit (20)

Sekhmet:
Limiting connections to 1 is a bit extreme, especially if running both indexer and API in the same instance.


bonustrack (approved):
utACK

bonustrack:
We may need to allow custom settings for this at some point. Snapshot hub currently use a connection limit of 1000.

PR #172, opened 2022-12-01
perf: skip DB query if requesting id only

Sekhmet:
## Summary

This PR will skip additional DB queries if only id is requested. New dependency had to be added to parse complex resolve info.

Closes https://github.com/snapshot-labs/checkpoint/issues/169

## Test plan

Change
```js
const item = await context.getLoader(returnType.name.toLowerCase()).load(id);
```

To
```js
console.log('querying', returnType.name.toLowerCase())
const item = await context.getLoader(returnType.name.toLowerCase()).load(id);
```

Run this request:
```gql
query {
  proposals {
    space {
      id
      strategies
    }
    author {
      id
    }
  }
}
```

You shouldn't see `querying User` in the console (only `querying Space`).


src/graphql/resolvers.ts
```diff
@@ -69,12 +73,27 @@ export async function queryMulti(parent, args, context: ResolverContext, info) {
...
   const jsonFields = getJsonFields(returnType);
 
   const id = parent?.[info.fieldName] || args.id;
 
+  const parsed = parseResolveInfo(info);
+  if (parsed) {
+    // @ts-ignore
+    const simplified = simplifyParsedResolveInfoFragmentWithType(parsed, returnType);
+
+    if (Object.keys(simplified.fields).length === 1 && simplified.fields['id']) {
```
bonustrack, line 92: What if "id" is not the primary key?

bonustrack: Actually I dont think we support different primary key so this may not be relevant

Sekhmet: It has to be `id` as we have convention that when using nested entity it will be linked to that entity `id` field.

bonustrack (approved):
utACK

PR #175, opened 2022-12-07
refactor: extract NetworkProvider

Sekhmet:
## Summary

This PR moves Starknet logic out of Checkpoint class into `NetworkProvider` that becomes responsible for fetching blocks and processing them. This allows for replacing indexing logic if needed.

I'm not sure if it's enough of abstraction to implement what we want, because I'm not sure how custom DB would work ("custom db (would be used for Snapshot)" from issue description. Would that query data by blockNum as well or it needs higher abstraction - if it does, some details would be helpful for context.

Closes https://github.com/snapshot-labs/checkpoint/issues/155

## Test plan

- Start `sx-api`.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -1,23 +1,16 @@
-import { RpcProvider } from 'starknet';
-import { validateAndParseAddress, hash } from 'starknet';
 import Promise from 'bluebird';
 import { addResolversToSchema } from '@graphql-tools/schema';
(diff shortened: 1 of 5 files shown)


bonustrack (approved):
@Sekhmet For custom DB like snapshot-hub a "block" would be more like a "batch" of for example 100 messages, the logic will be defined in the NetworkProvider class so your integration will works well with it.

tACK

PR #180, opened 2023-01-06
feat: parse events using specified format

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/checkpoint/issues/173

This PR adds `parseEvent` helper as well as `abi` option to Sources that can be used to automatically parse events that are available through `event`. Raw event is available through `rawEvent`.

**BREAKING CHANGE:** `event` param in writer is now reserved for parsed event, to access unparsed event use `rawEvent` instead.

## Test plan

Link `checkpoint` and apply to sx-api:
```diff
diff --git a/src/config.json b/src/config.json
index 453c2ea..d8ca574 100644
--- a/src/config.json
+++ b/src/config.json
@@ -1,8 +1,9 @@
 {
-  "network_node_url": "https://starknet-goerli.infura.io/v3/46a5dd9727bf48d4a132672d3f376146",
+  "network_node_url": "https://starknet-goerli2.infura.io/v3/46a5dd9727bf48d4a132672d3f376146",
   "sources": [
     {
       "contract": "0xe1e511e496a72791ab3d591ba7d571a32de4261d84e4d183f26b6325970e20",
+      "abi": "SpaceFactory",
       "start": 14808,
       "events": [
         {
diff --git a/src/index.ts b/src/index.ts
index 9d23fea..b10f572 100644
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,6 +5,7 @@ import path from 'path';
 import fs from 'fs';
 import Checkpoint, { LogLevel } from '@snapshot-labs/checkpoint';
 import config from './config.json';
+import spaceFactoryAbi from './spaceFactory.json';
 import * as writer from './writer';
 
 const dir = __dirname.endsWith('dist/src') ? '../' : '';
@@ -17,7 +18,10 @@ if (process.env.NETWORK_NODE_URL) {
 
 const checkpoint = new Checkpoint(config, writer, schema, {
   logLevel: LogLevel.Info,
-  prettifyLogs: process.env.NODE_ENV !== 'production'
+  prettifyLogs: process.env.NODE_ENV !== 'production',
+  abis: {
+    SpaceFactory: spaceFactoryAbi
+  }
 });
 
 checkpoint.reset().then(() => checkpoint.start());
diff --git a/src/writer.ts b/src/writer.ts
index 248bafc..2102ac1 100644
--- a/src/writer.ts
+++ b/src/writer.ts
@@ -12,6 +12,10 @@ function intSequenceToString(intSequence) {
     .join('');
 }
 
+function uint256toString(uint256) {
+  return (BigInt(uint256.low) + (BigInt(uint256.high) << BigInt(128))).toString();
+}
+
 export const handleSpaceCreated: CheckpointWriter = async ({
   block,
   tx,
@@ -21,24 +25,19 @@ export const handleSpaceCreated: CheckpointWriter = async ({
 }) => {
   if (!event) return;
 
-  console.log('Handle space created');
-  const format =
-    'deployer_address, space_address, voting_delay, min_voting_period, max_voting_period, proposal_threshold(uint256), controller, quorum(uint256), strategies_len, strategies(felt*), strategies_params_len, strategies_params(felt*), authenticators_len, authenticators(felt*), executors_len, executors(felt*)';
-  const data: any = getEvent(event.data, format);
-
   const item = {
-    id: validateAndParseAddress(data.space_address),
-    name: getSpaceName(data.space_address),
-    controller: validateAndParseAddress(data.controller),
-    voting_delay: BigInt(data.voting_delay).toString(),
-    min_voting_period: BigInt(data.min_voting_period).toString(),
-    max_voting_period: BigInt(data.max_voting_period).toString(),
-    proposal_threshold: data.proposal_threshold,
-    quorum: data.quorum,
-    strategies: JSON.stringify(data.strategies),
-    strategies_params: JSON.stringify(data.strategies_params),
-    authenticators: JSON.stringify(data.authenticators),
-    executors: JSON.stringify(data.executors),
+    id: validateAndParseAddress(event.space_address),
+    name: getSpaceName(event.space_address),
+    controller: validateAndParseAddress(event.controller),
+    voting_delay: BigInt(event.voting_delay).toString(),
+    min_voting_period: BigInt(event.min_voting_duration).toString(),
+    max_voting_period: BigInt(event.max_voting_duration).toString(),
+    proposal_threshold: uint256toString(event.proposal_threshold),
+    quorum: uint256toString(event.quorum),
+    strategies: JSON.stringify(event.strategies),
+    strategies_params: JSON.stringify(event.strategies_params),
+    authenticators: JSON.stringify(event.authenticators),
+    executors: JSON.stringify(event.executors),
     proposal_count: 0,
     vote_count: 0,
     created: block.timestamp,
@@ -54,15 +53,15 @@ export const handleSpaceCreated: CheckpointWriter = async ({
   await mysql.queryAsync(query, [item]);
 };
 
-export const handlePropose: CheckpointWriter = async ({ block, tx, event, mysql }) => {
-  if (!event) return;
+export const handlePropose: CheckpointWriter = async ({ block, tx, rawEvent, mysql }) => {
+  if (!rawEvent) return;
 
   console.log('Handle propose');
   const format =
     'proposal, author, quorum(uint256), timestamps, execution_strategy, execution_hash, metadata_uri_len, metadata_uri(felt*)';
-  const data: any = getEvent(event.data, format);
+  const data: any = getEvent(rawEvent.data, format);
 
-  const space = validateAndParseAddress(event.from_address);
+  const space = validateAndParseAddress(rawEvent.from_address);
   const [{ strategies, strategies_params }] = await mysql.queryAsync(
     'SELECT strategies, strategies_params FROM spaces WHERE id = ? LIMIT 1',
     [space]
@@ -135,14 +134,14 @@ export const handlePropose: CheckpointWriter = async ({ block, tx, event, mysql
   await mysql.queryAsync(query, [item, item.space, user, author]);
 };
 
-export const handleVote: CheckpointWriter = async ({ block, event, mysql }) => {
-  if (!event) return;
+export const handleVote: CheckpointWriter = async ({ block, rawEvent, mysql }) => {
+  if (!rawEvent) return;
 
   console.log('Handle vote');
   const format = 'proposal, voter, choice, vp';
-  const data: any = getEvent(event.data, format);
+  const data: any = getEvent(rawEvent.data, format);
 
-  const space = validateAndParseAddress(event.from_address);
+  const space = validateAndParseAddress(rawEvent.from_address);
   const proposal = parseInt(BigInt(data.proposal).toString());
   const voter = toAddress(data.voter);
   const choice = parseInt(BigInt(data.choice).toString());

```


src/utils/events.ts
```diff
@@ -0,0 +1,47 @@
+type Callback<Item> = Item extends { name: infer Name; type: infer Type }
```
bonustrack, line 1: This is specific to StarkNet, shouldn't this be stored at `src/providers/starknet/events.ts`?

src/index.ts
```diff
@@ -1,5 +1,6 @@
 import Checkpoint from './checkpoint';
 export { LogLevel } from './utils/logger';
+export { parseEvent } from './utils/events';
```
bonustrack, line 3: Any reason to expose this function at root level? Shouldn't this be exposed by StarkNet provider file?

bonustrack (commented):
Can be have also a more compact way to define type with a string like this:
```
deployer_address felt, space_address felt, voting_delay felt...
```
So we don't endup with too large config.json file hard to read. With Cairo 1 I believe the types will be standardized we may be able to use ABI file directly to get the format for all events.

Sekhmet:
> Can be have also a more compact way to define type with a string like this:
> > deployer_address felt, space_address felt, voting_delay felt...

What do you think about supporting both formats? The reason I went with this format is because you can just copy it straight from ABI and it will work, where with string format you would need to rewrite format from JSON.

We could make it accept either this object or string and it will use correct parser depending on what you pass.

bonustrack:
@Sekhmet Yes i think we should support both. How do you copy ABI when using types like "Address" which is a type only defined by us? Like this https://github.com/snapshot-labs/sx.js/blob/master/src/clients/starknet-tx/abi/space.json#L167

Sekhmet:
> How do you copy ABI when using types like "Address" which is a type only defined by us? Like this https://github.com/snapshot-labs/sx.js/blob/master/src/clients/starknet-tx/abi/space.json#L167

That is a good point. Do you think we should support adding custom types so you can still copy-paste them?

Something like this from starknet.py:
```py
uint256_abi = {
    "name": "Uint256",
    "type": "struct",
    "size": 2,
    "members": [
        {"name": "low", "offset": 0, "type": "felt"},
        {"name": "high", "offset": 1, "type": "felt"},
    ],
}

transfer_abi = {
    "name": "Transfer",
    "type": "event",
    "keys": [],
    "outputs": [
        {"name": "from_address", "type": "felt"},
        {"name": "to_address", "type": "felt"},
        {"name": "token_id", "type": "Uint256"},
    ],
}

transfer_decoder = FunctionCallSerializer(
    abi=transfer_abi,
    identifier_manager=identifier_manager_from_abi([
        transfer_abi, uint256_abi
    ]),
)
```

Without it support for JSON format would be mediocre.

bonustrack:
@Sekhmet Can we get custom types from the ABI directly? It seem to be exposed in the ABI https://github.com/snapshot-labs/sx.js/blob/master/src/clients/starknet-tx/abi/space.json#L56-L72
Technicaly we actually dont need to write the ABI again im the config, we could just have the ABI file in the instance and add `"contract": "Space"` in the config file

bonustrack:
> That's what I was suggesting. The types from my previous comment are from ABI. The issue with doing it automatically is that for example Uint256 would end up as { low, high } where it's way nicer if it returns BigInt automatically, maybe we could make exception for that one type.

I'm not sure if Uint256 type is a standard or used by others, if not maybe it's better that Checkpoint return { low, high } and we have a custom function in the writer to convert it. In future Cairo 1 would standardize such types.

> If we have something like this, do we still want deployer_address felt, space_address felt, voting_delay felt... or "abi" approach would be the only thing supported?

Agree, we don't need compact formar if we support ABI directly

bonustrack:
Wdyt if instead of having `event.parsed` we can return the parsed event as "event" and return the raw event as "eventRaw"?

Sekhmet:
I think it makes sense, the only issue I can see that it's not backwards compatible so it might be confusing during upgrades, but we can make sure it's explained in changelog.

bonustrack:
Sound good

bonustrack:
I've double checked that I run the correct version of CP but somehow `event` return the raw event for me, while `rawEvent` is `undefined`.

bonustrack:
I'll try with a reset of `node_modules` but I've added a `console.log` yesterday in my test and was able to see it

Sekhmet:
If you still get this issue after that, can you create branch of sx-api that you are testing against so I can try to reproduce?

bonustrack:
@Sekhmet Tried but got the same result, check PR here: https://github.com/snapshot-labs/sx-api/pull/162

bonustrack (approved):
tACK

PR #181, opened 2023-01-11
feat: allow for mysql ssl connections

janek26:
Allows for URLs like 
```
mysql://root:default_password@localhost:3306/checkpoint?sslaccept=strict
```
as used by in example planetscale


Sekhmet (approved):
utACK

PR #182, opened 2023-01-11
chore: run CI for pull requests from forks

Sekhmet:
## Summary

This way we can have tests trigger automatically if we have PR coming from a fork.


bonustrack:
@Sekhmet What if the fork expose our config vars?

Sekhmet:
I missed we are running publish on this workflow. Will update it to use 2 different workflows (one for release, only on master) and other for running tests (that don't have any secrets).

bonustrack (approved):
utACK

PR #184, opened 2023-01-13
feat: add support for optimistic indexing

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/checkpoint/issues/151

This PR adds support for optimistic indexing - now pending transactions are include when indexing. This option can be enabled with `optimistic_indexing` flag.

Assumptions:
- Writer will be called only once, either from pool or block, so writer needs to handle both scenarios if optimistic indexing is enabled.
- Pending transactions will be included in the next block (not +2 block)
- Pending transaction will always be included in the next block (can't fail/revert/disappear).

## Test plan

You can use this branch to test with: https://github.com/snapshot-labs/sx-api/tree/sekhmet/optimistic (remember to link this PR's checkpoint).
Not much difference is visible on goerli2 as it takes a while for transactions to show up in pending and block times are not that bad.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -7,6 +7,7 @@ import type { BaseProvider } from './providers';
 // Shortcuts to starknet types.
 export type Block = RPC.GetBlockWithTxs;
 export type Transaction = RPC.Transaction;
+export type PendingTransaction = RPC.PendingTransactions[number];
 export type Event = RPC.GetEventsResponse['events'][number];
 
 // (Partially) narrowed types as real types are not exported from `starknet`.
@@ -64,6 +65,7 @@ export type ContractTemplate = {
 // Configuration used to initialize Checkpoint
 export interface CheckpointConfig {
   network_node_url: string;
+  optimistic_indexing?: boolean;
   start?: number;
   tx_fn?: string;
   global_events?: ContractEventConfig[];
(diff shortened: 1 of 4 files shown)


bonustrack:
@Sekhmet Does this assume you see all txs that will be in the next block from the pending txs? If tx 0xabc doesn't appear as pending txs but appear on next block will it be going through writer? I don't see how you store the index of a transaction within a pending block, if the same pending block get updated multiple time with more txs how do you keep track of which txs were included and not?

src/types.ts
```diff
@@ -62,6 +63,7 @@ export type ContractTemplate = {
 // Configuration used to initialize Checkpoint
 export interface CheckpointConfig {
   network_node_url: string;
+  optimistic_indexing?: boolean;
   start?: number;
   tx_fn?: string;
   globalEvents?: ContractEventConfig[];
```
bonustrack, line 69: The casing seem wrong here by the way

bonustrack (commented):
@Sekhmet What happen if someone restart a Checkpoint instance (without reset) while some pending transaction(s) were added? From what I understand Checkpoint wont know where it stopped right? The blacklist seem to be only stored in vars on server memory.

Sekhmet:
> What happen if someone restart a Checkpoint instance (without reset) while some pending transaction(s) were added? From what I understand Checkpoint wont know where it stopped right? The blacklist seem to be only stored in vars on server memory.

That's a good point, we would need to store pending transactions in database then.

There is a slight chance than even without optimistic indexing we can still fire twice for event if server is restarted during events being processed (writer fired, block was not updated), we could re-use the same mechanism for this as for optimistic updates (processed transactions stored in DB), but there will be still 1 writer possible to fire twice if you get extremely lucky, which I think we can't do anything about unless we make writers use DB transactions instead of querying DB themselves.

bonustrack (approved):
utACK

PR #189, opened 2023-01-18
feat: add subentities support

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/checkpoint/issues/166

This PR adds `@derivedFrom` directive support that can be used to create 1-n queries.

## Test plan

Add a field on some entity that references other entity for example:
```gql
type Space {
  id: String!
  proposals: [Proposal]! @derivedFrom(field: "space")
}

type Proposal {
  id: String!
  space: Space
}
```

This will cause `proposals` to become available as part of Space schema resulting in this query to database (table name is derived from type used, column name is the field argument from directive):
```sql
SELECT * FROM proposals WHERE space IN (space.id)
```

Now such queries can be used:
```gql
{
  spaces {
    id
    proposals {
      id
      tx
      title
      strategies
    }
  }
}
```

You can test it against this branch on sx-api (extra changes due to event parsing): https://github.com/snapshot-labs/sx-api/tree/sekhmet/subentities

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -39,8 +39,9 @@ export default class Checkpoint {
   ) {
     this.config = config;
     this.writer = writer;
-    this.schema = schema;
-    this.entityController = new GqlEntityController(schema, opts);
+    this.schema = this.extendSchema(schema);
+
+    this.entityController = new GqlEntityController(this.schema, opts);
 
     this.sourceContracts = getContractsFromConfig(config);
     this.cpBlocksCache = [];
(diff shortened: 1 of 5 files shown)


bonustrack (approved):
tACK. If we always return all the entries it may take too long to resolve or become unusable, ideally we should limit the results and allow custom query like root operation example:
```gql
{ 
  spaces {
    name
    proposals (first: 10, orderBy: end) {
      title
    }
  }
}
```
But this can be solved in future, this PR is already a great step.

PR #190, opened 2023-01-19
feat: allow ABI to be specified in templates

Sekhmet:
This way we can use automatic parsing of events for template-initiated contracts.

## Test plan

Can be tested with https://github.com/snapshot-labs/sx-api/pull/163


bonustrack (approved):
utACK

PR #191, opened 2023-01-19
refactor: rename globalEvents to global_events

Sekhmet:
## Summary

Renaming `globalEvents` to `global_events` to match other config options.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -209,7 +209,7 @@ export default class Checkpoint {
 
     const nextBlock = lastBlock + 1;
 
-    if (this.config.tx_fn || this.config.globalEvents) {
+    if (this.config.tx_fn || this.config.global_events) {
       if (this.config.start) start = this.config.start;
     } else {
       (this.config.sources || []).forEach(source => {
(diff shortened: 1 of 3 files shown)


bonustrack:
How do you define ABI for global events?

Sekhmet:
@bonustrack it's not possible, I'm not sure if it should be possible - event names are generic so there might not be a single ABI for event named XYZ, so I imagine people working with those might want to handle those themselves, check if event is valid etc. but maybe it makes sense to still provide option to define ABI per event.

bonustrack:
The events for ERC20 are standardized not only the functions, some contract may have more functions but the core ones like events name and types should always be the same, following this:  https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/token/erc20/library.cairo

bonustrack:
Right, let's keep it without ABI for now. Do we return the event in `rawEvent`?

Sekhmet:
Yes: https://github.com/snapshot-labs/checkpoint/blob/e61e967b5cf4590448c0b991cc364b0a87881477/src/providers/starknet/index.ts#L96

bonustrack (approved):
utACK

PR #193, opened 2023-01-26
fix: pass local index when passing nested structs

Sekhmet:
## Summary

When parsing nested structs we should pass index for current index within the struct. Previously it would pass index where its parent starts.

## Test plan

Test case makes sense (it used to return `{ high: '0x0', low: '0x23cdb7dc255bf4da' }` because it was reading choice as part of Uint256.


bonustrack (approved):
utACK

PR #195, opened 2023-01-31
fix: add special pluralization for metadata

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/checkpoint/issues/194

This PR adds special pluralization for metadata so it becomes querable again.
Alternatively we could use default pluralization (for database, still need it for GraphQL), but it will mean that DB schema will be different.

## Test plan

Run this query:
```gql
{
  _metadatas {
    id
    value
  }
  _metadata(id: "last_indexed_block") {
    id
    value
  }
}
```

### DIFF
--- a/src/graphql/index.ts
+++ b/src/graphql/index.ts
@@ -8,8 +8,8 @@ import {
   GraphQLString
 } from 'graphql';
 import DataLoader from 'dataloader';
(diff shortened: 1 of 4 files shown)


bonustrack (approved):
tACK

PR #196, opened 2023-02-02
feat: add config validation (abis, writers)

Sekhmet:
## Summary

This PR adds validation on config:
- Checking if all used Abis were passed
- Checking if all used writers are defined

## Test plan

- Rename writer in config/writers file. It should complain.
- Rename abi in config/opts. It should complain.


bonustrack (approved):
utACK

PR #207, opened 2023-03-15
refactor: use knex as query builder for other DB engines support

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/checkpoint/issues/202

This PR replaces `mysql` specific internal code to use `knex` that makes it compatible with MySQL/MariaDB and PostgreSQL (and possibly SQLite3 + MSSQL, but those were not tested fully).

Why:
- using DB independent query builder allows us to support other DB engines (most notably PostgreSQL).
- using `knex` means that we get this interoperability out-of-the-box, with interface that is already well documented and popular in the ecosystem (in the future this access will most likely replaced by ORM, where knex is used internally).
 
 
 ## Test plan
 
 - Build this PR.
 - Follow this PR's testplan to try sx-api with PostgreSQL: https://github.com/snapshot-labs/sx-api/pull/179


package.json
```diff
@@ -36,13 +36,15 @@
...
     "@types/bn.js": "^5.1.0",
     "@types/jest": "^27.4.1",
-    "@types/mysql": "^2.15.21",
     "@types/node": "^18.11.6",
     "dotenv": "^16.0.1",
     "eslint": "^8.28.0",
     "jest": "^27.5.1",
     "jest-mock-extended": "^2.0.7",
+    "mysql": "^2.18.1",
+    "pg": "^8.10.0",
```
Sekhmet, line 45: Users will need to install DB client locally, we are installing those only as dev-dependencies to make those work when linking packages.

bonustrack (approved):
tACK works well now. What do you think if we not use a file path as env for the certificate but a string directly (otherwise I imagine the certificate would need to public in the repo)

Sekhmet:
> What do you think if we not use a file path as env for the certificate but a string directly (otherwise I imagine the certificate would need to public in the repo)

I now added `CA_CERT` (for string) and `CA_CERT_FILE` (for file path).


Tests are failing because for some reason Github Actions uses partially old files when building :thinking:

PR #209, opened 2023-03-16
feat: log processing errors

Sekhmet:
Currently if errors occurs when processing blocks there is no feedback (other than seeing indexer process same block over and over).
We should log those errors so it's easy to see the issues.

## Test plan
- Test with https://github.com/snapshot-labs/sx-api/pull/177 by linking checkpoint
- Break something in the code, for example change `spaces` table name to `borked` in query to insert new proposal).
- Run `yarn dev` on sx-api.
- Errors get logged.


bonustrack (approved):
tACK

PR #210, opened 2023-03-16
feat: add resetMetadata function

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/checkpoint/issues/197

If you end up with bad state (you added new event you want to index) there is currently no way of resetting it via checkpoint and you need to delete tables manually from SQL.

This function will handle truncating Checkpoint's metadata, it will be later used to automatically handle those cases.

## Test plan

Test with: https://github.com/snapshot-labs/sx-api/pull/177

Run 

```ts
async function run() {
  await checkpoint.reset();
  await checkpoint.resetMetadata();
  await checkpoint.start();
}

run();
```

It starts from zero.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -145,6 +145,19 @@ export default class Checkpoint {
     await this.entityController.createEntityStores(this.mysql);
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
utACK

PR #211, opened 2023-03-20
fix: convert resetStore to use knex

Sekhmet:
Tests were actually failing for good reason, but I didn't pull latest master so I looked like CI issue.

This PR removes `mysql` usage in checkpoint's store that ended up there after merging two last PRs.


bonustrack (approved):
utACK

PR #213, opened 2023-03-20
feat: add ORM Model

Sekhmet:
## Summary

Depends on https://github.com/snapshot-labs/checkpoint/pull/212

This PR adds simple Model class that will be used as base for all generated models from schema.

What is missing:
 - codegen to generate models from  GraphQL schemas (with this we probably could start make it public).
 - transactions - all writes from writers should be batched into database transaction, together with call that updates current blockNumber so database stays consistent
 - rollback support - with all requests going through ORM we can store data in DB that will allow us to revert inserts/update/deletes in case of rollbacks.
 
 ## Test plan
 
 - Run this code: https://github.com/snapshot-labs/sx-api/compare/sekhmet/orm?expand=1
 - Run `users` GraphQL query to make sure it works.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -1,24 +1,24 @@
 import Promise from 'bluebird';
+import { GraphQLObjectType, GraphQLSchema } from 'graphql';
 import { addResolversToSchema } from '@graphql-tools/schema';
 import { Knex } from 'knex';
 import { Pool as PgPool } from 'pg';
 import getGraphQL, { CheckpointsGraphQLObject, MetadataGraphQLObject } from './graphql';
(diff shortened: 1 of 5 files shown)


bonustrack:
@Sekhmet Just tested and it works well, the format and approach you have is great. I don't think we need rollback support just yet, there is different ways to do that, it can be done later.

Sekhmet:
@bonustrack 

Do you think we can merge this? It doesn't need to be exposed, but we can start building other parts of ORM support on top of it.

bonustrack (approved):
@Sekhmet Yes feel free to merge

Issue #185, opened 2022-12-19
bug: proposal counter is wrong


bonustrack:
Proposal counter is showing 1 proposal less than the actual number, see:
<img width="1001" alt="image" src="https://user-images.githubusercontent.com/16245250/208442473-2e0c1c0b-40b3-4366-a23e-79ed34fc4e3d.png">

bonustrack:
@Sekhmet How could we query block event that is not in sync? That would be a weird bug from StarkNet node. Do you think this happen only when we sync a newly created block? Would be nice if we can reproduce or find a way so we can keep track of what happened. I can see currently latest proposal on SX UI got the id 58, but there is only 56 proposals available.

Sekhmet:
I will try to reproduce it to get more data.

Sekhmet:
@bonustrack I was able to reproduce it.

It's indeed missing events returned from API. It's not a Starknet bug, but the way Infura (or other providers) might be serving nodes. In most cases they will have multiple nodes and you will get one at random when you make a request.
It might be possible that one of them is slightly behind so when node A sees a block and we make a request for events at the same block if we get node B to serve this request it might be behind and it will return empty array.

This should only happen when we are syncing live, past blocks shouldn't have this issue (unless Infura exposes node that is 1000 blocks behind).

As a fix I was thinking of doing this:
1. Fetch block N and events from block N.
2. If block exists but events are empty refetch later after a delay (as-if block wasn't there).
3. Fetch block N and events from block N again: if events are there just continue processing the block. If events are not there after 3 such attempts assume that block has no events (is this possible case, it looks like it might be possible) just continue with empty events.

As alternative we could use RPC batches, but those are not supported in starknet.js so we would need to make requests manually. This will allow us to make those two requests at once, but I'm not sure if it would guarantee solving this issue - RPC providers might still have some proxy that will separate the batch and still send them to two different nodes.

bonustrack:
@Sekhmet Can't we know directly with the request to get events if the block is available regardless of the block having events or not? If you query events from block in future (that doesnt exists) it will look the same as a previous block without events?

Sekhmet:
If you use blocks in the future it returns empty array (we don't query single block, but a range of blocks as a filter, so this might by why it does that instead of returning error).

PR #217, opened 2023-03-30
feat: add additional checks if there are no events in the block

Sekhmet:
## Summary

Closes: https://github.com/checkpoint-labs/checkpoint/issues/185

This is a workaround for issue that occurs when requesting block and events data from two different nodes, which might result in events being empty, due to second node not knowing about that block yet.

With this workaround empty events response will be reject, unless it shows up as such for 3 consecutive ticks (currently 7s each), will be reduced to 2s in upcoming PR.

Virtually all new blocks have events, so this is unlikely to trigger when syncing current data unless it's really an issue.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -280,7 +280,7 @@ export default class Checkpoint {
         } catch (err) {
           this.log.error({ blockNumber: blockNum, err }, 'error occured during pool processing');
         }
-      } else {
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
utACK

Sekhmet:
Replaced by https://github.com/checkpoint-labs/checkpoint/pull/234

PR #218, opened 2023-03-30
feat: add fetchInterval option

Sekhmet:
## Summary

This PR adds fetchInterval option to change default interval.


bonustrack (approved):
utACK

Sekhmet:
@bonustrack I can rebase it not to depend on https://github.com/checkpoint-labs/checkpoint/pull/217, but then we would also increase default interval back to 7 seconds, because this https://github.com/checkpoint-labs/checkpoint/issues/185 is still an issue. Would that work?

bonustrack:
@Sekhmet Yes

PR #219, opened 2023-03-31
fix: allow calling orderBy without orderDirection

Sekhmet:
After knex refactor it's now called with toLowerCase() so it has to be defined.


bonustrack (approved):
utACK

Issue #198, opened 2023-02-03
feat: detect checkpoints/metadata being outdated


Sekhmet:
## Summary

Currently there are cases that can cause checkpoint not to reset gracefully if some changes in config are made:
1. you change `start` property
2. you add new event
3. you change network https://github.com/checkpoint-labs/checkpoint/pull/221

This results in mixed state, where:
1. and 2. checkpoint won't go back to parse past event
3. checkpoint will have old state and possibly get stuck if number of blocks between networks differs

We can allow devs to trigger reset manually (https://github.com/snapshot-labs/checkpoint/issues/197), but it would be also great to run some checks to attempt to fix that automatically (or at least detect and suggest changes).

perfectmak:
So a solution I'm thinking to this will be to compute a checksum of the json configuration on start. If this is the first run, this will be saved in the metadata table, and on subsequent runs, this latest configurations checksum will be compared against the one in the database, and if they don't match, the index will fail to start.

Then to allow users to be able to resolve this efficiently, I'm thinking of adding a parameter to the `start` and `reset` functions that users can optionally pass to the function to force start even though there is a change in schema or in the case of reset to clean up the checksum in the config metadata's table.

What is your thought on this @Sekhmet?

perfectmak:
After testing the implementation described above, I realized that it would be a better experience to not just fail with a generic error message, but should also include some hint as to what could have changed. So I'm also going to try and store some extra information to provide suggestions on what has changed.

Sekhmet:
Sounds good to me, few quick thoughts, that you might want to consider:
1. ideally checksum should not care about slight changes of a file (formatting, order or fields etc.), maybe it should look only at block number (can store it as-is) and events checksum.
2. it would be cool if changing `network_node_url` wouldn't result in Checkpoint failing to start. Sometimes you just change it to different provider, doesn't make sense to make it fail automatically. Maybe we should ping `starknet_chainId` and detect changing chainId instead of just the URL, this way we don't break if someone just moves from Infura to Chainstack.
3. DX would be nice if we could have separate error messages for those cases (block changed, network changed, events were updated).

perfectmak:
Those a great suggestions. I'm using object-hash to generate the hash and it is based on an object containing specified fields extracted from the config object, subtle changes in the file should cause an issue.

But yeah, I'm modifying my changes to give correct helpful error messages.

Sekhmet:
[...]

PR #221, opened 2023-03-31
feat: add consistent network verification

Sekhmet:
Towards https://github.com/checkpoint-labs/checkpoint/issues/198

This PR adds verification that network returned by provider is the same as previously used.

## Test plan

- Start the server normally so network gets persisted.
- Change `network_node_url` in your config to different network.


bonustrack (approved):
tACK

PR #223, opened 2023-04-25
feat: continue fetching receipts on error

solimander:
Currently, `processPool` is aborted if a single `getTransactionReceipt` throws. Instead, process the receipts that do get returned. It's quite common for a pending transaction receipt to be missing when hitting a cluster of nodes.

### DIFF
--- a/src/providers/starknet/index.ts
+++ b/src/providers/starknet/index.ts
@@ -62,7 +62,7 @@ export class StarknetProvider extends BaseProvider {
         try {
           return await this.provider.getTransactionReceipt(tx.transaction_hash);
         } catch (err) {
-          this.log.error(
+          this.log.warn(
             { transactionHash: tx.transaction_hash, err },
             'getting transaction receipt failed'
           );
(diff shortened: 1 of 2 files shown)


solimander:
@Sekhmet Ah, thanks! Missed that the original transactions array is still processed. I'll update the solution to filter transactions that do not yet have a receipt.

solimander:
All set 👍

src/providers/starknet/index.ts
```diff
@@ -59,18 +59,27 @@ export class StarknetProvider extends BaseProvider {
       txs.map(async tx => {
         if (!tx.transaction_hash) return null;
 
-        return this.provider.getTransactionReceipt(tx.transaction_hash);
+        try {
+          return await this.provider.getTransactionReceipt(tx.transaction_hash);
+        } catch (err) {
+          this.log.error(
```
Sekhmet, line 65: Now it's not really an error case, as we kind of expect it to happen and handle it. Maybe it should be `.warn` or `.info` instead?

Sekhmet (approved):
utACK

Thanks!

PR #227, opened 2023-05-15
feat: detect start block changes

Sekhmet:
Depends on: https://github.com/checkpoint-labs/checkpoint/pull/221

This PR adds verification that start block in config did not change.

## Test plan

- Start the server normally so start block gets persisted.
- Change `start` in your config to different network (if editing events it has to be the lowest block number from there).


bonustrack (approved):
tACK

bonustrack (changes requested):
Actually something is wrong, even when I use reset() it give me the error:
image

Sekhmet:
@bonustrack reset only resets entities tables, you need to reset both (`reset()` and `resetMetadata()` - it is mentioned in error message). We have two separate methods, because resetMetadata was added later.

bonustrack (approved):

PR #228, opened 2023-05-15
feat: detect events and generic config changes

Sekhmet:
Depends on: https://github.com/checkpoint-labs/checkpoint/pull/232
Closes: https://github.com/checkpoint-labs/checkpoint/issues/198

This PR adds verification that tracked events didn't change.

## Test plan

- Start the server normally so start checksum gets persisted.
- Do some change in the config that you expect that should be detected (writer/abi names are not considered as they are user defined).

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "json-to-graphql-query": "^2.2.4",
     "knex": "^2.4.2",
     "mysql": "^2.18.1",
+    "object-hash": "^3.0.0",
     "pg": "^8.10.0",
     "pino": "^8.3.1",
     "pino-pretty": "^8.1.0",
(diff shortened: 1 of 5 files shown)


bonustrack:
Tried but `resetMetadata` doesn't work for some reason, I can see the table is not cleared 
image

Sekhmet:
Good catch, it was unrelated to those PRs, submitted PR with a fix and rebased this one on top of it to make it testable: https://github.com/checkpoint-labs/checkpoint/pull/232

bonustrack:
Just tested and noticed that when I remove an event, it doesn't get detected, it only show the error message if I add a new event.

Sekhmet:
@bonustrack I couldn't reproduce this issue directly, but I've found different bug, can you check if you still have this issue?
If yes, make sure you are not calling `resetMetadata` or don't have `resetOnConfigChange` set.

bonustrack:
It works for me now

bonustrack (approved):
tACK

Issue #205, opened 2023-03-15
feat: create a simple ORM 


bonustrack:
We should create a simple ORM, this will help insert data without writing SQL query, support different database or rewrite queries for support feature like time-travel queries and reorgs support. The ORM could work in similar way than TheGraph with this format:

```ts
const proposal = new Proposal(id);
proposal.author = '0x12345...';
proposal.start = 1234567890;
proposal.space = space;

await proposal.save();
```

To do so we would need to have a script to generate the types, we can start a `checkpoint` cli for this. 

The ORM would only handle `INSERT` for the moment, doing `SELECT` queries is not part of this issue, it will be spec in another one.

PR #229, opened 2023-05-16
feat: add codegen to generate ORM models from schema

Sekhmet:
Closes: https://github.com/checkpoint-labs/checkpoint/issues/205
Depends on: https://github.com/checkpoint-labs/checkpoint/pull/230

This PR adds very primitive codegen for our ORM. This is exposed with `checkpoint codegen` script.

**BREAKING CHANGE: I had to move `decimalTypes` from runtime opts to CheckpointConfig as it needs to be defined statically so we can use it in scripts.**

## Test plan
- Build checkpoint.
- Go to some repo (for example [sx-api](https://github.com/snapshot-labs/sx-api/compare/sekhmet/codegen?expand=1) - this branch is converted to use ORM) that uses checkpoint and link checkpoint.
- Run `node path/to/checkpoint/dist/src/bin/index.js generate`.
- `.checkpoint/models.ts` should be generated.
- Try using models to interact with DB - if using sx-api from above repo is already converted to use this ORM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "pino": "^8.3.1",
     "pino-pretty": "^8.1.0",
     "pluralize": "^8.0.0",
-    "starknet": "^4.13.1"
+    "starknet": "^4.13.1",
+    "yargs": "^17.7.2"
   },
   "devDependencies": {
     "@snapshot-labs/eslint-config": "^0.1.0-beta.7",
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "@snapshot-labs/checkpoint",
   "version": "0.1.0-beta.17",
   "license": "MIT",
+  "bin": {
+    "checkpoint": "dist/src/bin/index.js"
+  },
   "main": "dist/src/index.js",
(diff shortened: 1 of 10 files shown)


bonustrack:
Works well! Can't find any issue when testing. Some improvement I imagine would be to use a MySQL/Postgres transaction to batch insert but I assume it can come later. I just have a small request, can we rename the command to from `codegen` to `generate` and change the folder from `codegen` to `.checkpoint`? This is similar than what Prisma is doing.

Sekhmet:
> Some improvement I imagine would be to use a MySQL/Postgres transaction to batch insert but I assume it can come later.

Yeah, it's one of the improvements that I planned there: https://github.com/checkpoint-labs/checkpoint/pull/213 - maybe we should create new issues out of it to track those.

Sekhmet:
@bonustrack this should be valid JS (ESM) file, should we target CJS instead?

bonustrack:
Ok got it, let's stick with ESM, we would add CJS if there is some integration that requires it.

bonustrack (approved):
tACK

PR #230, opened 2023-05-18
chore: upgrade node to 18

Sekhmet:
Node 14 is out of cycle already, and 16 is about to be dropped in couple months, we can move to 18 which is current LTS.

Upgrading jest because our current version doesn't work with node 18 that I'm using locally.


bonustrack:
I would recommend to wait for this DigitalOcean dont support v18

bonustrack:
Just to confirm this would just force v18 for the test right? Cuz DO still doesn't support v18

Sekhmet:
Yes, I also tested installing deps with v16 and v14 in case any new packages force specific version, but both work still.

PR #231, opened 2023-05-18
feat: receive next block number from processBlock

bonustrack:
This add the logic to define next block to sync in the provider directly within the `processBlock` function. This helps in some use case I'm testing where I sync a database which doesn't have blocks, in my setup a block is a transaction. But to sync the db it's not efficient to do it transaction per transaction I want to use batch of transaction instead. The issue is that if I use batch when I get close to have the db fully sync, I can not predict what will be the number of transaction available in the batch, by adding the next block within the provider `processBlock` function I can just return the last transaction order id to make sure the sync continue at the right transaction. There might be better approach to it but I found this to work well.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -272,9 +272,9 @@ export default class Checkpoint {
     this.log.debug({ blockNumber: blockNum }, 'next block');
 
     try {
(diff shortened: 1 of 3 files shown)


Sekhmet (approved):
tACK

PR #232, opened 2023-05-23
fix: truncate Metadata table

Sekhmet:
Due to error we were truncating checkpoints table twice instead of checkpoints table and metadata table.


bonustrack (approved):
utACK

Issue #200, opened 2023-02-09
feat: reduce delay between loading block retries


bonustrack:
We should reduce the delay for retry loading a block if the block does not exists, here: https://github.com/snapshot-labs/checkpoint/blob/master/src/checkpoint.ts#L240
We can reduce the delay to 2sec, this will help on the UI to see changes in realtime. Ideally this should be part of the options in Checkpoint.

Issue #168, opened 2022-11-24
feat: quick first indexing


bonustrack:
~~One idea to speedup indexing on first sync or when adding new contract would be to create a server using Checkpoint which index all the checkpoints for all the contracts on StarkNet to allow server running Checkpoint instance to seed checkpoints from it.~~

Another approach would be to use the provider function `getEvents` for multiple blocks at once to find checkpoints and seed them into the db

bonustrack:
Another approach would be to use the provider function `getEvents` for multiple blocks at once to find checkpoints and seed them into the db

Sekhmet:
> Another approach would be to use the provider function getEvents for multiple blocks at once to find checkpoints and seed them into the db

For this, you think we should checkpoint all events, even those unrelated for the indexer (because we won't know which ones are related).

This might end up with lots of data to store, there are currently 3.5 million events on goerli2, 82 million on goerli1, if we assume 32 bytes per record (address is 32 bytes on its own, ignoring block number to adjust for possible compression), this means:
goerli2: 112 GB
goerli1: 2.6 TB

This is for full history, so it's unlikely that most applications will need to store this much, but it doesn't scale to have all apps store it locally.

This also won't help with global_events tracking as we would need event_hash/blockNumber checkpoints then, which will increase size significantly too if we handle that.

anhlam2234:
> ~~One idea to speedup indexing on first sync or when adding new contract would be to create a server using Checkpoint which index all the checkpoints for all the contracts on StarkNet to allow server running Checkpoint instance to seed checkpoints from it.~~
> 
> 
> 
> Another approach would be to use the provider function `getEvents` for multiple blocks at once to find checkpoints and seed them into the db

PR #234, opened 2023-06-01
feat: add prefetch for blocks

Sekhmet:
Closes: https://github.com/checkpoint-labs/checkpoint/issues/168
Closes: https://github.com/checkpoint-labs/checkpoint/issues/185
Closes: https://github.com/checkpoint-labs/checkpoint/issues/200

This PR adds prefetch for events, making it possible to only fetch blocks we know we have blocks for. Prefetch will speed up catching up to latest block (-50) by fetching events from 100 blocks at the time, and storing them in database, from where they are queried by normal indexing process to find events from source contracts. Old unused events are later purged.

## Test plan

You can test it by running it as usual, no extra changes on consumer side should be needed - it should just be much faster at finding new events.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -323,7 +323,7 @@ export default class Checkpoint {
   }
 
   private async nextEvents(blockNum: number) {
-    if (!this.prefetchEndBlock) throw new Error('preloadEndBlock is not set');
+    if (!this.prefetchEndBlock) throw new Error('prefetchEndBlock is not set');
     if (blockNum > this.prefetchEndBlock) {
       this.prefetchDone = true;
       return;
@@ -344,7 +344,7 @@ export default class Checkpoint {
   }
 
   private async next(blockNum: number) {
-    if (!this.prefetchEndBlock) throw new Error('preloadEndBlock is not set');
+    if (!this.prefetchEndBlock) throw new Error('prefetchEndBlock is not set');
 
     if (!this.config.tx_fn && !this.config.global_events) {
       if (blockNum <= this.prefetchEndBlock) {
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -25,7 +25,7 @@ import {
 
 const BLOCK_PRELOAD = 100;
 const BLOCK_PRELOAD_OFFSET = 50;
-const DEFAULT_FETCH_INTERVAL = 7000;
+const DEFAULT_FETCH_INTERVAL = 2000;
 
 export default class Checkpoint {
   public config: CheckpointConfig;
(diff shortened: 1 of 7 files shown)


bonustrack:
Did some test, does the flow work this way: first you index all events, then you run the indexing of the contract with writers? Or is it supposed to work in parallel? Somehow I can't get any event indexed in the db with `sx-api`, writers are not triggered.

Sekhmet:
@bonustrack it should work in parallel, I'm testing against sx-api and have no issues.

Make sure to use `reset` to address the issue mentioned in PR description. If it still doesn't work can you share the logs?

bonustrack:
It seem that now the indexing only start once there are checkpoints available, what would be the behavior when an instance is fully sync and we stream new block 1 by 1? Shouldn't prefetch stop at this point? I'm concerned with prefetch it will slow down indexing new events as we will wait for prefetch process to bring new event before we can index them.

bonustrack:
@Sekhmet Got it thanks, this address my concern. I'm thinking also there is some usecase where you don't need prefetch, for example when I use CP to sync a db, I can query only the events that I'm interested in and prefetch is useless, would be great to add an option to being able to disable prefetch for this case.

bonustrack:
@antiyro Would be great if you can test this PR too

Sekhmet:
> Got it thanks, this address my concern. I'm thinking also there is some usecase where you don't need prefetch, for example when I use CP to sync a db, I can query only the events that I'm interested in and prefetch is useless, would be great to add an option to being able to disable prefetch for this case.

Can you show example of this usage? Hard to come up with some solution without seeing it. I assume with this approach you don't need checkpoints at all?

bonustrack:
Here is same example code I'm using to sync a db with CP https://github.com/snapshot-labs/api-checkpoint it indeed not need checkpoints here as I want to sync everything.

bonustrack:
Got this behavior not sure if it's intended, it seem that a function running twice with same block number
image

Sekhmet:
@bonustrack it's intended, we have main loop that does the syncing and it will check if there are any prefetched checkpoints   and if not it will sleep for 10 seconds and check later.

bonustrack:
@Sekhmet Ok and this can happen even if the instance is not fully sync right?

bonustrack:
Also once it's fully sync I see `handling pool` requests, is this the normal behavior? Shouldn't prefetch stop once fully sync?:
image

Sekhmet:
> Ok and this can happen even if the instance is not fully sync right?

This will only happen when it's not fully synced, after that it will disable prefetch.

> Also once it's fully sync I see handling pool requests, is this the normal behavior? Shouldn't prefetch stop once fully sync?:

Pool requests are for optimistic indexing, so those happen when fully synced. "checkpoint inserted" is for prefetch and only happens if not fully synced.

bonustrack (approved):
tACK

PR #236, opened 2023-06-05
feat: persist templates

Sekhmet:
We need to persist templates, otherwise when restarting the server it will continue processing blocks from last indexed block, but with information about executed templates it's likely to omit events.

## Test plan
- Start sx-api for ~100 blocks.
- Log `this.config.sources` here: https://github.com/checkpoint-labs/checkpoint/blob/c5e4deb22462b05b509e917277d11b8e884e988d/src/checkpoint.ts#L309
- Start sx-api without calling .reset (just .start())
- It should log both factory and one space in sources.


bonustrack (approved):
Works well, you may want to update controller.ts to remove "CheckpointOptions" otherwise the lint script return a warning

Issue #11, opened 2022-04-21
feat: validate Checkpoint config


perfectmak:
Currently, Checkpoint assumes that the config object passed into the constructor is properly structured, but this will not always be the case. 
Validation should be done on the config objects structure before starting the indexer. Maybe a using json schema validation.

ef3n9r98:
Hey @perfectmak, I know some devs who'd be interested in working on this. Is there a budget for fixing this issue, and if so, how much? 

I'd put this issue up on [OpenQ]([www.openq.dev](http://www.openq.dev/)).

Thanks!

PR #237, opened 2023-06-06
feat: validate checkpoint config with zod

Sekhmet:
Closes: https://github.com/checkpoint-labs/checkpoint/issues/11

## Test plan
- Start sx-api - it works
- Do some change in config.json, remove non-optional property, change type of value (add `// @ts-ignore` to make it compile).
- Start sx-api, you get error

### DIFF
--- a/src/schemas.ts
+++ b/src/schemas.ts
@@ -1,0 +1,37 @@
+import { z } from 'zod';
+
+export const contractEventConfigSchema = z.object({
+  name: z.string(),
+  fn: z.string()
+});
+
+export const contractSourceConfigSchema = z.object({
+  contract: z.string(),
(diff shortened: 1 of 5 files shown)


src/schemas.ts
```diff
@@ -0,0 +1,37 @@
...
+  events: z.array(contractEventConfigSchema)
+});
+
+export const contractTemplateSchema = z.object({
+  abi: z.string().optional(),
+  events: z.array(contractEventConfigSchema)
+});
+
+export const checkpointConfigSchema = z.object({
+  network_node_url: z.string(),
```
bonustrack, line 22: Could we check if it's an url too?

bonustrack (approved):
tACK

PR #244, opened 2023-07-06
feat: auto-increment for entity id

0xTitan:
Hello.
Implementation done. Auto increment annotation can be added only once on entity definition. Only Int, BigInt and ID (for nested object) are managed.

Eg :
`scalar Text

type Post {
  id: Int! @autoIncrement
  author: String!
  content: Text!
  tag: String
  tx_hash: String!
  created_at: Int!
  created_at_block: Int!
  poster: Poster
}

type Poster {
  id: ID! @autoIncrement
  name: String!
}`

Unfortunately i'm struggling with knex implementation and the primaryKey definition. There is normally the possiblity to define if an autoIncrement field should be defined as primaryKey, this is not working when i'm setting it false. Due to that i can only managed one autoIncrement field per table. Maybe enought for now.

Eg : 

[Knex increment](https://knexjs.org/guide/schema-builder.html#increments)

`
knex.schema.createTable('users', function (table) {
  table.increments('id');
  table.increments('other_id', { primaryKey: false });
});
`

I added two more tests :

- should work with autoincrement
- should work with autoincrement and nested objects

image

I put the constant needed under type.ts. Let me know if it's suitable.

The result under myssql for the schema above :
image


Sekhmet (changes requested):
`@autoincrement` can be handled as native [GraphQL directive](https://spec.graphql.org/October2021/#sec-Language.Directives), like [`@derivedFrom`](https://github.com/checkpoint-labs/checkpoint/blob/master/src/utils/graphql.ts#L12)

0xTitan:
Ok thx, will check that.

0xTitan:
Update done. New directive @autoIncrement added.

src/graphql/controller.ts
```diff
@@ -205,34 +219,79 @@ export class GqlEntityController {
...
-
     return { builder };
   }
 
+  /**
+   * Parse schema to extract table and fields witrh annotation autoIncrement
+   * @param schema
+   * @returns A map where key is table name and values a list of fields with annotation autoIncrement
+   */
+  private extractAutoIncrementFields(schema: string) {
```
Sekhmet, line 264: This shouldn't be handled with regexes. Directive can be handled natively with GraphQL.

Sekhmet: As we extended GraphQL with our custom directive [here](https://github.com/checkpoint-labs/checkpoint/blob/23698a4587b4fe2b1160fe0224401e0d5d1bb90f/src/utils/graphql.ts#L12-L16), we can then access that directive directly from our GraphQL tree.

This way you can look at the field (id field in this example, and check if it has `@autoincrement` directive, same as we do with `@derivedFrom`:
https://github.com/checkpoint-labs/checkpoint/blob/e6dd8b3580604d39de79f4c8c41fb2103582677c/src/graphql/resolvers.ts#L182-L186

That means we don't have to bother with writing our own dummy parser for GraphQL to extract autoIncrementFields, and just check for `field.astNode.directives` here:
https://github.com/checkpoint-labs/checkpoint/blob/23698a4587b4fe2b1160fe0224401e0d5d1bb90f/src/graphql/controller.ts#L223-L228

0xTitan: OK thx for the hint. I will check it and modify accordingly.

Sekhmet:
@0xTitan thank you for your contribution so far. Are you still expecting to work on this issue? If not I could push it across finish line.

0xTitan:
Hi @Sekhmet, i pushed changes to remove regex and use directive. Huge thanks for your help and your patience !

Sekhmet (commented):
Looks way cleaner thank you!

General thoughts:
- can you revert any unrelated changes to the formatting? Most of it would be solved by running Prettier or `yarn lint`.
- the comments added are a bit too much in my opinion as they explain pretty self-explanatory code, so they could be removed
- we have added ORM since and it won't work well with it (as currently ORM requires you to create new instances of ORM models by passing ID to a constructor) - is this something you are also willing to take a look at? Otherwise we probably should at least prevent ORM models from being created for entities with autoincrement fields until it's implemented. If you don't want to work on that I could at this logic so we could get your PR merged.

src/graphql/controller.ts
```diff
@@ -212,31 +224,46 @@ export class GqlEntityController {
...
-        t.primary(['id']);
+        //if there is no autoIncrement fields on current table, mark the id as primary
+        const tableHasAutoIncrement = this.getTypeFields(type).some(field => {
+          const directives = field.astNode?.directives ?? [];
+          const autoIncrementDirective = directives.find(dir => dir.name.value === 'autoIncrement');
+          return autoIncrementDirective;
+        })
+
+        if (!tableHasAutoIncrement)
+          t.primary(['id']);
```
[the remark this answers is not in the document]
0xTitan: Thanks a lot for the review. I updated code based on your comments.
Regarding ORM, i could take a look, does the ORM is available on the prisma branch ?

package.json
```diff
@@ -61,4 +61,4 @@
     "dist/**/*",
     "src/**/*"
   ]
-}
+}
```
Sekhmet, line 64: Can you revert this change?

src/graphql/controller.ts
```diff
@@ -212,31 +211,41 @@ export class GqlEntityController {
...
-              ? t[sqlType.name](field.name, ...sqlType.options)
-              : t[sqlType.name](field.name);
 
-          if (field.type instanceof GraphQLNonNull) {
-            column = column.notNullable();
-          }
+          const directives = field.astNode?.directives ?? [];
+          const autoIncrementDirective = directives.find(dir => dir.name.value === 'autoIncrement');
+
+          //Check if field is declared as autoincrement
```
Sekhmet, line 223:
```suggestion
```

test/unit/graphql/controller.test.ts
```diff
@@ -69,10 +69,43 @@ type Vote {
...
+
+      const controller = new GqlEntityController(schema);
       const { builder } = await controller.createEntityStores(mockKnex);
 
-      expect(builder.toString()).toMatchSnapshot();
+      const createQuery = builder.toString();
+      expect(createQuery).toMatchSnapshot();
+    });
+
+    it('should work with autoincrement and nested objects', async () => {
```
Sekhmet, line 81: I feel like it could just be the only test case in this file (`it should work`).

0xTitan:
Thanks for your comments. Update done.

Sekhmet:
Sorry about that, but with new setup where entities have versions it doesn't really make sense (each version would autoincrement and entities won't have stable ID).

I think it could be technically possible to do but since the code changes so much (with historical versions, ORM) it would be much more complex and I'm not sure there is a good use case to warrant this added complexity.

PR #245, opened 2023-08-24
feat: parse Cairo1 events

Sekhmet:
This PR adds support for parsing events with Cairo1 ABIs.

Tests should pass, it tests both Cairo0 and Cairo1 events.


bonustrack (approved):
utACK

PR #246, opened 2023-08-29
feat: process new sources events in same transaction

Sekhmet:
Previously sources added at transaction A would not be processed at that transaction, only when next transaction is processed.

This PR changes this behaviour causing new sources to be processed immediatelly, within the same transaction they were created (assuming start block expects this).


bonustrack (approved):
utACK

PR #247, opened 2023-09-05
feat: add nested entity filters

Sekhmet:
Depends on https://github.com/checkpoint-labs/checkpoint/pull/246

This PR adds support for filtering entities by their subentities, for example following query becomes now possible:
```gql
{
  spaces (where: { metadata_: { name_contains: "Shadow DAO"}}) {
    id
    metadata {
      id
      name
      executors
      executors_types
    }
    voting_delay
  }
}
```

## Test plan

Test against https://github.com/snapshot-labs/sx-api/pull/181

### DIFF
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -49,6 +49,15 @@ const GraphQLOrderDirection = new GraphQLEnumType({
   }
 });
 
+type WhereResult = {
+  where: {
+    type: GraphQLInputObjectType;
+  };
+  orderByValues: Record<string, { value: string }>;
+};
+
+const cache = new Map<string, WhereResult>();
+
 /**
  * Controller for performing actions based on the graphql schema provided to its
  * constructor. It exposes public functions to generate graphql or database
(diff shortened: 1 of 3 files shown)


src/graphql/controller.ts
```diff
@@ -297,78 +306,96 @@ export class GqlEntityController {
     type: GraphQLObjectType,
     resolver: GraphQLFieldResolver<Parent, Context>
   ): GraphQLFieldConfig<Parent, Context> {
-    const whereInputConfig: GraphQLInputObjectTypeConfig = {
-      name: `Where${type.name}`,
-      fields: {}
-    };
+    const getWhereType = (type: GraphQLObjectType<any, any>): WhereResult => {
+      const name = `${type.name}_filter`;
```
Sekhmet, line 310: Normalized filter name to match TheGraph.

bonustrack:
Awesome, this is a great feature. I tried and it seem to work well, I have just these 2 issues:

When I try to query the sub field "id" it will fail with this error:
image
I've tested on TheGraph a similar query and it works (they keep "id" filter in the sub entity "where").

Also when I try to query a sub entity using a required field as filter it fail. Here is my schema:
```gql
...
type User {
  id: String!
  name: String
  about: Text
  discussion_count: Int! # required field
  vote_count: Int! # required field
}
...
```
Here is the response I get:
image

Without this feature I imagine we wouldn't be able to do things like search on proposals on SX-UI, now I understand why you were looking on adding metadata directly in the Space entity level. It's great that you take that different approach, it will make it easier to integrate on UI as we can keep same format with the subgraph APIs.

bonustrack (approved):
tACK

PR #252, opened 2023-09-20
feat: increase ID length to 256

Sekhmet:
Vote IDs do not fit as they have 2 Starknet addresses + proposal ID


bonustrack (approved):
utACK

PR #253, opened 2023-09-20
fix: handle non-nullable List types

Sekhmet:
Allows querying fields like this
```
authenticators: String[]!
```

Can be tested with https://github.com/snapshot-labs/sx-api/pull/186


bonustrack (approved):
utACK

Issue #248, opened 2023-09-12
feat: allow for integer id


bonustrack:
Currently we can only use a String as id, we should allow integer too

PR #255, opened 2023-09-25
feat: support both String and Int entity IDs

Sekhmet:
Closes: https://github.com/checkpoint-labs/checkpoint/issues/248

## Test plan

- Change your id to `Int!` and run codegen with `node ../checkpoint/dist/src/bin/index.js generate`.

### DIFF
--- a/src/codegen.ts
+++ b/src/codegen.ts
@@ -131,9 +131,25 @@ export const codegen = (
     contents += `export class ${modelName} extends Model {\n`;
     contents += `  static tableName = '${pluralize(modelName.toLowerCase())}';\n\n`;
 
-    contents += format === 'javascript' ? `  constructor(id) {\n` : `  constructor(id: string) {\n`;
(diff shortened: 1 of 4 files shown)


bonustrack (commented):
It fail when I start the server with this error:
```
https://github.com/checkpoint-labs/checkpoint/blob/2d66accfed49db4fa43d0a55dc3bf111f7249922/src/graphql/controller.ts#L504
```
And this message:
> Error: sql type for Category not support

It seem to that entity type are not working. I'm testing with this branch & schema https://github.com/snapshot-labs/highlight/blob/fabien/int-id/src/api/schema.gql 

This might not be easy to run locally it uses combine Highlight with Redis + Checkpoint.

Sekhmet:
I see, it happens if you use such entity with Int id field in relations. Will work on that.

bonustrack (approved):
Works well now tACK

PR #256, opened 2023-09-25
fix: encode JSON fields on load

Sekhmet:
JSON fields have to be encoded when saving on PostgreSQL. We already encode them when set manually, but don't when we get those values from database on load.

## Test plan
- Use [this PR](https://github.com/snapshot-labs/sx-api/pull/186) with PostgreSQL.
- Link and regenerate models with  `node ../checkpoint/dist/src/bin/index.js generate`. (you need to do it with bin file directly, `yarn checkpoint generate` won't work.


bonustrack (approved):
tACK

Issue #249, opened 2023-09-12
bug: ORM return null for 0 or false value


bonustrack:
I got a value in a row that is `0`, but when I try to return it on a writer it give me `null`, the issue is here:
https://github.com/checkpoint-labs/checkpoint/blob/9a9eb462fca95336e2554954a108840c0b422a58/src/orm/model.ts#L40
```ts
const category = await Category.loadEntity(id);

if (category) {
    console.log(category.parent); // return null instead of 0
}
```

PR #257, opened 2023-09-25
fix: do not return null for all falsy values

Sekhmet:
Closes: https://github.com/checkpoint-labs/checkpoint/issues/249


bonustrack (approved):
utACK

PR #259, opened 2023-09-27
fix: restore cpBlocksCache if processing failed

Sekhmet:
Currently it's possible that processing fails and block won't be retried because it was in cpBlocksCache and got removed before processing. We need to restore it back.

## Test plan
- Throw in some handler (for example handleVote).
- Run server.
- It keeps retrying instead of skipping.


bonustrack (approved):
utACK

Issue #258, opened 2023-09-25
bug: Query with nested where clause fails with "Cannot read properties of undefined (reading 'toLowerCase')"


solimander:
Queries with nested where clauses fail [here](https://github.com/checkpoint-labs/checkpoint/blob/e6dd8b3580604d39de79f4c8c41fb2103582677c/src/graphql/resolvers.ts#L74).

Looks like line 73 should be:

```typescript
const nestedReturnType = (
  returnType.getFields()[fieldName].type as GraphQLList<GraphQLObjectType>
).ofType;
````

I don't mind opening a PR for this, but want to double check this can be reproduced and isn't some misconfiguration on my end.

**Comment by bonustrack** [comment]
@solimander Can you share the schema and query that you are using to reproduce this issue?

**Comment by solimander** [comment]
Sure thing. Here's the [schema](https://github.com/Prop-House/prop-house-monorepo/blob/25cc8e1df04f4e769c26e7c4dd28876b28fd8813/packages/prop-house-checkpoint/src/schema.gql).

Query:

```gql
query {
  votes(where: { round_: { sourceChainRound: "round_address" }, proposal_: { proposalId: 1 } }) {
    id
  }
}
```

PR #261, opened 2023-09-27
fix: make sure data from where clause is complete

Sekhmet:
Closes: https://github.com/checkpoint-labs/checkpoint/issues/258

This PR fixes two issues:
- When querying non-nullable nested field in where prevents crash.
- When using nested where resolve objects and lists of scalars.

## Test plan
Run query like this on sx-api:
```gql
{
  votes(where: {space_: {}}) {
    space {
      id
      metadata {
        id
        name
        executors
      }
    }
  }
}
```

### DIFF
--- a/src/graphql/resolvers.ts
+++ b/src/graphql/resolvers.ts
@@ -6,6 +6,7 @@ import {
   GraphQLResolveInfo,
   GraphQLScalarType,
   isListType,
+  isObjectType,
   isScalarType
 } from 'graphql';
 import {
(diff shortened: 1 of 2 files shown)


bonustrack:
@solimander Does this PR fix the issue you get?

bonustrack (approved):
tACK

solimander:
> @solimander Does this PR fix the issue you get?

Yes, looks like that solves it! 🙏

PR #265, opened 2023-10-02
feat: rename OrderBy type name to match TheGraph

Sekhmet:
We need to make it compatible with TheGraph so we can use single query across Checkpoint and TheGraph.


bonustrack (approved):
utACK

PR #268, opened 2023-10-05
fix: ensure subentities are defined in where type

Sekhmet:
Closes: https://github.com/checkpoint-labs/checkpoint/issues/267

### DIFF
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -331,7 +331,7 @@ export class GqlEntityController {
             idField &&
             idField.type instanceof GraphQLNonNull &&
             idField.type.ofType instanceof GraphQLScalarType &&
-            ['String', 'ID'].includes(idField.type.ofType.name)
+            ['String', 'Int', 'ID'].includes(idField.type.ofType.name)
           ) {
             if (level === 0) {
               whereInputConfig.fields[`${field.name}_`] = getWhereType(
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -56,8 +56,6 @@ type WhereResult = {
   orderByValues: Record<string, { value: string }>;
 };
 
-const cache = new Map<string, WhereResult>();
-
 /**
  * Controller for performing actions based on the graphql schema provided to its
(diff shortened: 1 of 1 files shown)


bonustrack:
It works now with `Int` as id for sub-entity but somehow now `String` as id for sub-entity doesn't work anymore, where filters are not created for the sub-entity, see my query: 
image

```gql
query {
  users  (where: { profile_: { id: "0x4C7909d6F029b3a5798143C843F4f8e5341a3473" }}) {
    id
    profile {
      id
      discord
    }
  }
}
```

Also query would fail if there is an entry without sub-entity (this might be the normal behavior):
image

```gql
query {
  topics {
    id
    category {
      metadata_uri
    }
  }
}
```

This is using same schema here: https://github.com/snapshot-labs/highlight/blob/main/src/api/schema.gql

Sekhmet:
@bonustrack can you test this? I believe it will use different type name for nested queries than TheGraph, but it will work for sx-api purposes - I think they support multi-level nested queries which we don't, so we have to use unique names for where type with subentities, and without subentities (for example Topic_filter has subentities, but Topic_User_filter won't).

bonustrack:
This did the fix for the first query. For the 2nd query I got this error:
```
"The loader.load() function must be called with a value, but got: undefined.
```
And the 3rd I got
```
Cannot return null for non-nullable field Category.id.
```

Sekhmet:
What do you expect to happen in this case if database state is inconsistent (which I assume if the case here) - you reference object that doesn't exist in the database?

Can you do DB dump so I can reproduce this?

Sekhmet:
@bonustrack tackled the issue with missing entities in different PR as those are unrelated:
https://github.com/checkpoint-labs/checkpoint/pull/269

If you set `category` to null it will return null. If you set `category` but subentity is missing it will return null, but will also add error about missing entity.

I couldn't find a way to reproduce loader error. Do you know when it happens?

bonustrack (approved):
tACK

bonustrack (approved):

PR #269, opened 2023-10-06
fix: handle nested entity id set to null

Sekhmet:
Depends on: https://github.com/checkpoint-labs/checkpoint/pull/268
Addresses: https://github.com/checkpoint-labs/checkpoint/pull/268#issuecomment-1748812601

If nullable entity type's ID field is set to null we should return null on entity.

I'm not sure if it fixed issue with `The loader.load() function must be called with a value, but got: undefined.` as I couldn't find a way to reproduce.


bonustrack (approved):
tACK, I was initially using `0` as index on the db instead of `null` which wasn't working even with this PR but moved to `null` and with this PR it works now

PR #277, opened 2024-01-23
fix: select checkpoint blocks by distinct

Sekhmet:
If there are multiple contracts checkpointed at same block, this block can then be processed multiple times causing issues.


bonustrack (approved):
utACK, is this to fix the issue with SX API?

Sekhmet:
Yes, it is.

PR #279, opened 2024-02-26
fix: use indexes on checkpoint table

Sekhmet:
Those need to be indexed otherwise optimistic indexing will become slow.


bonustrack (approved):
utACK

PR #282, opened 2024-03-15
feat: prevent processing same transaction twice

Sekhmet:
We've seen cases of pool transactions being processed multiple times. It's possible to happen if RPC node returns outdated transactions from pool. We only cache seen pool transactions for single block, so if new block is minted but RPC returns pool before that block it could trigger twice.

Now we cache all processed transactions - this should be managable memory wise and prevent it from ever happening.


bonustrack (approved):
utACK

PR #284, opened 2024-04-02
fix: prevent fetching pool transactions every tick

Sekhmet:
Currently we fetch receipt of each transaction that is in pending block on every tick, even if we already saw this transaction before.

This is causing lots of requests as block can have ~400 transactions sitting in it for minutes and we hit it every 2 seconds.

Now instead of fetching those and skipping this in handlePool we skip those before they are even fetched.


bonustrack (approved):
utACK 🤑

PR #286, opened 2024-04-11
fix: start prefetch on last seeded block

Sekhmet:
Currently when using seed checkpoints main loop will start syncing from those seeded blocks, but optimistic loop will prefetch block starting with star block, ignoring all seeds.

This PR addresses it so LastPrefetchedBlock will be adjusted when `seedCheckpoints` is called.


bonustrack (approved):
utACK

PR #287, opened 2024-04-12
feat: add support for non-derived one-to-many relations

Sekhmet:
Currently you can define one-to-many relations using @derivedFrom directive. Using it the field on the entity is just virtual field that gets filled in at query time using lookup on the derived entity's column.

This PR adds support for non-derived one-to-many relations - this field needs to be filled in by writers to be usable. You specify ID's of referenced entity and those will be filled in with full object at query time.

### DIFF
--- a/src/codegen.ts
+++ b/src/codegen.ts
@@ -13,7 +13,7 @@ import {
 } from 'graphql';
 import pluralize from 'pluralize';
 import { GqlEntityController } from './graphql/controller';
-import { extendSchema } from './utils/graphql';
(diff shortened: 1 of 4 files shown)


bonustrack:
Would that make it possible to query a subquery like this? 
```gql
query {
  space(id: "0x123") {
      id
      proposals (where: { vote_count_gt: 10 }}) {
        id
      }
  }
}

Sekhmet:
Not at the moment, although I could add that I think. This was added to support executions on Starknet which didn't require filters like this.

This is unrelated to this PR though.

bonustrack (approved):
utACK

PR #290, opened 2024-04-19
fix: resolve inner join columns with non-derived list fields

Sekhmet:
Follow up to: https://github.com/checkpoint-labs/checkpoint/pull/287

Forgot to commit this change before merging.


bonustrack (approved):
utACK

PR #292, opened 2024-04-26
feat: prefetch blocks by targetting contracts

Sekhmet:
Closes: https://github.com/snapshot-labs/pitches/issues/75

This PR replaces old prefetch with new system that will load only events from contracts we are interested in.

This makes initial sync much faster (5 minutes instead of couple hours).

### Test plan

- Link this PR in `sx-monorepo` (also link `graphql` from `node_modules/graphql`) and build.
- Previous step might not work right away, if you get missing packages error when starting server nuke `node_modules` in both repos, reinstall and relink again - super inconsistent.
- Add `checkpoint.resetMetadata` before `reset` call.
- Start sx-api `NETWORK_NODE_URL=https://starknet-sepolia.infura.io/v3/AAAA NETWORK=SN_SEPOLIA yarn dev`
- Once synced (should take couple of minutes) check responses for some requests (for example all spaces with proposal/vote counts) and compare it with https://testnet-api-1.snapshotx.xyz/

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -23,7 +23,7 @@ import {
   TemplateSource
 } from './types';
 
-const BLOCK_PRELOAD = 100;
+const BLOCK_PRELOAD = 1000;
 const BLOCK_PRELOAD_OFFSET = 50;
 const DEFAULT_FETCH_INTERVAL = 2000;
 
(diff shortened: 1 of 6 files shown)


bonustrack (approved):
tACK

Issue #61, opened 2022-06-16
feat: lookback query at specific block number


bonustrack:
Currently when you query Checkpoint API you get only data sync from the latest state of the chain. Some apps like SX-UI will need to query the API with state at a specific block number in the past for resolving things like delegation for vote.
One way we could enable this lookback feature is to store all mutations with their corresponding block number on the db. If we want to query data at latest state of the chain we would just query with MySQL only the entries with the highest block number; And if we query with a state at a specific block number we would query with MySQL just the entries with a block number before or equal to the defined state block number.

One issue with this approach is that we can't enforce how data is stored on Checkpoint so it would be hard for dev to understand how their MySQL INSERT queries should look like, we would probably need to have some kind of in-house ORM to handle insert queries, it could be simple like:
```ts
const vote = { proposal: 1, choice: 2 };
await db.insert('Vote', vote);
```

perfectmak:
First, I'm going to agree that Checkpoint needs some ORM, so we can abstract some logic and optimisations that will potentially be useful to its users (for example, it will be helpful to implement [this feature](https://github.com/snapshot-labs/checkpoint/issues/10) as well).

Now, concerning tracking mutations and being able to query the state at a block. I think for this to work correctly and be helpful; we will need to track more than inserts; we will also need to track updates and deletions of entities. Thinking more about the database structure and the logic to fetch entities at a specific block height will help determine this.

Here is are my thoughts on how we can implement this:

# Tracking Entity State Plan

We will need to keep track of each entity's state after every mutation. This plan is taking some inspiration from how The Graph [currently does](https://github.com/graphprotocol/graph-node/issues/813#issuecomment-473115438). We can achieve this by tracking the block range for which an entity's state is valid and only creating a new entity record with an updated block range when there is a change to the entity.

I'll be using the `Vote` entity in your example to continue my description and assume it has a schema like:

```graphql
type Vote {
  id: String!
  proposal: Int!
  choice: Int!
}
```

Currently, we generate an SQL table of similar structure to the entities graphql field, BUT if we want to track entity states at specific block heights, we should create a table for each entity like this:

| Column | Type |
| --- | --- |
| id | VARCHAR NOT NULL |
| proposal | INT NOT NULL |
| choice | INT NOT NULL |
| _block_min | INT NOT NULL |
| _block_max | INT NULLABLE |

[...]

PR #297, opened 2024-06-11
feat: add time travel queries

Sekhmet:
Closes: https://github.com/checkpoint-labs/checkpoint/issues/61

This PR adds support for time travel queries by storing mutations of entities with ranges those are valid for.
You can specify `block` filter` when querying API to get version at given block.

This will be also used for reorg support to be able to revert changes.

Changes:
- MySQL support is removed to use PostgreSQL's native `int8range` type.
- Direct `pg` access in writers is removed - without knowing the exact inner-workings of our data structure queries on SQL level will most likely result in different effect that user will expect. All queries should be done through ORM that encapsulates all voodoo we have to do.
- `bluebird` was removed (wanted to remove it earlier, but we used it to promisify mysql, now it was removed entirely and only usage was replaced by simple `sleep` helper. Now this gives us better type-safety.
- `block` filter is available for single and multi entity queries.

### Examples

Test with this branch which has more complex schema to test all edge-cases: https://github.com/checkpoint-labs/checkpoint-template/compare/sekhmet/tt?expand=1

```gql
{
  users (block: 65309) {
    id
    created_at
    created_at_block
    posts {
      id
      content
      created_at_block
      tags {
        id
      }
    }
    info {
      post_count
    }
  }
}
```


src/checkpoint.ts
```diff
@@ -299,12 +296,11 @@ export default class Checkpoint {
   public async getWriterParams(): Promise<{
     instance: Checkpoint;
     knex: Knex;
-    mysql: AsyncMySqlPool;
     pg: PgPool;
   }> {
     return {
       instance: this,
-      mysql: this.mysql,
+      knex: this.knex,
```
Sekhmet, line 303: We should remove knex and pg from there as well.

bonustrack:
Did some small test and it seem to work well. I noticed using `block: 0` would return all the results, I don't think that should be the case but it's just a detail.

bonustrack (approved):
tACK, tested with the branch `sekhmet/tt` from Checkpoint template adding some entity that get updated and so far its working well

```gql
type Overview {
  id: String!
  post_count: Int!
}
```

Sekhmet:
@bonustrack I updated checkpoint-template to use more complex scenario that should cover all cases (with sample query in this block post).

Also handled `0` block and added schema version verification so database will be reset once this version gets introduced automatically (because otherwise database will be in broken state).

If it all still looks good to you I will merge this PR, do release and migrate our APIs to this version.

bonustrack:
Also does it support delete entity?

Sekhmet:
> Also does it support delete entity?

Yes, you can delete entities.

> Looks good, does this version works for subqueries assuming range?

Not sure if I understand it, but `block` you specify affects all subentities (so for example in sample query when you request `User` with `UserInfo` at specific block that `UserInfo` will be version for the block you requested).

Issue #300, opened 2024-06-25
Bug: last_indexed_block reset to Zero after reaching the seed blocks


shardAstronaut:
### Description
I'm indexing ETH contract on Starknet, the indexer index all the blocks but just after reaching last seed block the `last_indexed_block`  is set to **ZERO** and the data is deleted.

```json
"@snapshot-labs/checkpoint": "^0.1.0-beta.31",
```

![issue](https://github.com/checkpoint-labs/checkpoint/assets/127413534/0c231d0e-caeb-47e5-8e73-204cf28e194c)

bonustrack:
@shardAstronaut This is weird, it might have been fixed already. Can you try using a more recent version of Checkpoint? `0.1.0-beta.35` or `0.1.0-beta.36` (this latest version support only Postgres and Checkpoint ORM).

shardAstronaut:
@bonustrack I upgraded the package to beta.36 and I still have the same issue. On the logs I get this error `{"error":"RangeError: Maximum call stack size exceeded"}`
I'll indexing without seed blocks and see.

bonustrack:
@shardAstronaut Is your code available on GitHub to check?

shardAstronaut:
@bonustrack 
I pushed the code to this repo https://github.com/shardAstronaut/checkpoint-template
When running the indexer you will get this log error just after the indexer reach the seed blocks.
```sh
[1719415002170] DEBUG: handling transaction done
    component: "checkpoint"
    txIndex: 141
[1719415002170] DEBUG: handling transaction
    component: "checkpoint"
    txIndex: 142
[1719415002170] INFO: found contract event
    component: "checkpoint"
    contract: "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
    event: "Transfer"
    handlerFn: "liquity_eth_Transfer"
[1719415002171] DEBUG: handling transaction done
    component: "checkpoint"
    txIndex: 142
[1719415002171] DEBUG: handling transaction
    component: "checkpoint"
    txIndex: 143
[1719415002171] INFO: found contract event
    component: "checkpoint"
    contract: "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
    event: "Transfer"
    handlerFn: "liquity_eth_Transfer"
[1719415002171] INFO: found contract event
    component: "checkpoint"
[...]

PR #302, opened 2024-06-26
fix: Maximum call stack size exceeded with array.push

shardAstronaut:
### Description

- fix: Maximum call stack size exceeded with array.push (Bug: last_indexed_block reset to Zero after reaching the seed blocks #300)

When indexing contracts with a big number of events the array.push method will throw an error.
When you use spread operator all items of the source array are stored in the stack as arguments list, so having a large number of items (~ > 100K) will cause the this stack size exceed. The most simple work-around is to manually push all items one by one.
https://stackoverflow.com/questions/61740599/rangeerror-maximum-call-stack-size-exceeded-with-array-push

### DIFF
--- a/src/providers/starknet/provider.ts
+++ b/src/providers/starknet/provider.ts
@@ -217,7 +217,7 @@ export class StarknetProvider extends BaseProvider {
     }
 
     let lastSources = this.instance.getCurrentSources(blockNumber);
-    const sourcesQueue = [...lastSources];
+    let sourcesQueue = [...lastSources];
 
     let source: ContractSourceConfig | undefined;
     while ((source = sourcesQueue.shift())) {
@@ -299,7 +299,7 @@ export class StarknetProvider extends BaseProvider {
         nextSource => !lastSources.find(lastSource => lastSource.contract === nextSource.contract)
       );
 
-      sourcesQueue.push(...newSources);
+      sourcesQueue = sourcesQueue.concat(newSources);
       lastSources = nextSources;
     }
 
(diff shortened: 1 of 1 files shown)


Sekhmet (commented):
Do you know if this issue still happens in your case if we use concat instead of `push` in a loop?

I wonder if `concat`:
1. solves the issue
2. is faster

shardAstronaut:
> Do you know if this issue still happens in your case if we use concat instead of `push` in a loop?
> 
> I wonder if `concat`:
> 
> 1. solves the issue
> 2. is faster

Looks like concat is faster than `push` based on this discussion https://stackoverflow.com/questions/48865710/spread-operator-vs-array-concat.
I can update the PR to use concat it's cleaner, wdyt?

Sekhmet:
Let's do concat then, thanks!

shardAstronaut:
Do you know what changed in Starknet sequancer to start getting this error now? I was indexing the ETH contract for more than 4 months and never get this error.

Sekhmet:
> Do you know what changed in Starknet sequancer to start getting this error now? I was indexing the ETH contract for more than 4 months and never get this error.

Most likely it's caused by the fact that we changed the way we fetch events to find blocks we need to index (before we fetched all events, but only for 100 blocks at the time, now we fetch 1000 blocks or more at the time, but filtering events that only come from contracts we track). Generally it causes way faster indexing, but looks like there is room for improvements for certain contracts.

Sekhmet (approved):
utACK

Sekhmet:
Thank you!

Issue #301, opened 2024-06-25
bug: _checkpoints entity can not be queried anymore


bonustrack:
There is now an error to query `_metadatas` from the GraphQL API, this was most likely introduced in #297  
```
"select \"_metadatas\".* from \"_metadatas\" where upper_inf(_metadatas.block_range) limit $1 - column _metadatas.block_range does not exist"
```
<img width="1056" alt="image" src="https://github.com/checkpoint-labs/checkpoint/assets/16245250/ff5df937-67b0-4ff8-a46b-df90dae0feb5">

PR #303, opened 2024-06-26
fix: do not apply block range filter on internal tables

Sekhmet:
This commit also refactors out applyBlockFilter helper into database utils to avoid repetition.

Closes: https://github.com/checkpoint-labs/checkpoint/issues/301

### Test plan

Run query like this:
```gql
query ($checkpointId: ID!, $metadataId: ID!) {
  spaces(block: 18407) {
    id
  }
  _checkpoints {
    id
  }
  _metadatas {
    id
  }
  _checkpoint(id: $checkpointId) {
    id
  }
  _metadata(id: $metadataId) {
    id
  }
}
```
It works without crashing.

### DIFF
--- a/src/graphql/index.ts
+++ b/src/graphql/index.ts
@@ -9,7 +9,7 @@ import {
 } from 'graphql';
 import DataLoader from 'dataloader';
 import { ResolverContextInput } from './resolvers';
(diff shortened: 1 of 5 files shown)


bonustrack (approved):
utACK

PR #304, opened 2024-06-27
fix: deduplicate checkpoints

Sekhmet:
If there are multiple events in single block we should only add one entry
into preloadBlocks. Otherwise it will cause `next` to be called multiple times
for the same block.


bonustrack (approved):
utACK

Issue #308, opened 2024-07-01
bug: handling block ends up taking 1 by 1 block even where no events


bonustrack:
When starting to sync an instance, the handling block iterative function starts taking 5 to 10 blocks, then minutes later it just sync block 1 by 1, even if there is no event matching, this slowing down indexing. We should only go through block with events.

<img width="495" alt="image" src="https://github.com/checkpoint-labs/checkpoint/assets/16245250/c4ba38b5-0efb-44ae-9fff-4010bc2503d0">

Sekhmet:
Normally, prefetch ends when we reach latest block - 100 (at the time of checkpoint startup). It looks like it's not this case though as those blocks are from 2 weeks ago.

Do you have repro for this? I didn't notice it happening.

bonustrack:
@Sekhmet It was using this code: https://github.com/snapshot-labs/delegates-api/tree/70874e92167433e8ca3fdafecafc120e22a6e564

bonustrack:
It seem that every block with an event from config contract get indexed, even if the specific event happening is not listed in config.

PR #310, opened 2024-07-03
feat: filter starknet events by event names

Sekhmet:
Closes: https://github.com/checkpoint-labs/checkpoint/issues/308

In case of contracts that have lots of activity overall, but we only care about few events we should filter events by keys as well. This way we are not processing blocks that do not contain blocks that we are interested with.

## Test plan

1. Test with https://github.com/snapshot-labs/delegates-api/commits/fabien/multi-tokens/
2. It should be quick to sync (I only tested with NSTR token so it's quicker, currently testing with STRK, but it has lots of activity we are targeting so it won't be much faster).

### DIFF
--- a/src/providers/starknet/provider.ts
+++ b/src/providers/starknet/provider.ts
@@ -343,7 +343,8 @@ export class StarknetProvider extends BaseProvider {
   async getCheckpointsRangeForAddress(
     fromBlock: number,
     toBlock: number,
-    address: string
+    address: string,
+    eventNames: string[]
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
Awesome, works well now it's much faster few min vs hours tACK

bonustrack:
Do we have the same on EVM?

Sekhmet:
Haven't looked for solution for EVM yet, will need proper tests for that, just wanted to have workaround for Starknet so delegates API can get synced.

PR #314, opened 2024-09-11
fix: prevent errors in preload from crashing app

Sekhmet:
Currently if getCheckpointsRange throws in preload it will cause app to crash.

We need to handle this and retry.


bonustrack (approved):
utACK

PR #316, opened 2024-10-09
fix: prevent duplicate entries with overlapping block_ranges

Sekhmet:
It was possible to create duplicate entires that would apply within the same block that would have single ID.

That shouldn't be the case, ID is not longer unique in whole table, but should be within single block_range.

### DIFF
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -205,6 +205,10 @@ export class GqlEntityController {
   public async createEntityStores(knex: Knex): Promise<{ builder: Knex.SchemaBuilder }> {
     let builder = knex.schema;
 
+    if (knex.client.config.client === 'pg') {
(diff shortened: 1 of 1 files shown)


src/graphql/controller.ts
```diff
@@ -205,6 +205,10 @@ export class GqlEntityController {
   public async createEntityStores(knex: Knex): Promise<{ builder: Knex.SchemaBuilder }> {
     let builder = knex.schema;
 
+    if (knex.client.config.client === 'pg') {
```
bonustrack, line 208: Do we support other db than Postgres? Or what is this condition for?

Sekhmet: At the moment just for our tests as those run with SQLite.

bonustrack (approved):
utACK

PR #318, opened 2024-11-01
fix: maximum call stack in evm provider

pscott:
Similarly to https://github.com/checkpoint-labs/checkpoint/pull/302/files , fix the call stack size issue with `evm` provider

### DIFF
--- a/src/providers/evm/provider.ts
+++ b/src/providers/evm/provider.ts
@@ -151,7 +151,7 @@ export class EvmProvider extends BaseProvider {
     }
 
     let lastSources = this.instance.getCurrentSources(blockNumber);
-    const sourcesQueue = [...lastSources];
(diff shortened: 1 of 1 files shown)


src/providers/evm/provider.ts
```diff
@@ -217,7 +217,7 @@ export class EvmProvider extends BaseProvider {
     return events.reduce((acc, event) => {
       if (!acc[event.transactionHash]) acc[event.transactionHash] = [];
 
-      acc[event.transactionHash].push(event);
+      acc[event.transactionHash] = acc[event.transactionHash].concat(event);
```
pscott, line 220: I think that's what we need to do but I'd like you to double check

Sekhmet (approved):
utACK

PR #319, opened 2024-11-05
feat: add filters for list types

Sekhmet:
## Summary

This commit adds support for filters for list of scalar types:
- list
- list_not
- list_contains
- list_not_contains

Those filters are implemented using in PostgreSQL and only available on JSONB so we changed column type for those fields.

This will allow us to implement https://github.com/snapshot-labs/workflow/issues/254 again.

## Test plan:

Run sx-api on Sepolia with default start index for a bit to get some data (1 minute should be enough), run following queries:

```gql
{
  spaces(
    where: {
      authenticators: [
        "0x2b63a8a92b7c67484ab99c4307c7db41b15b9a3c33359cd2b2459fd7f543a9c"
      ]
    }
  ) {
    id
    authenticators
  }
}
```

```gql
{
  spaces(
    where: {
      authenticators_not: [
        "0x2b63a8a92b7c67484ab99c4307c7db41b15b9a3c33359cd2b2459fd7f543a9c"
      ]
    }
  ) {
    id
    authenticators
  }
}
```

```gql
{
  spaces(
    where: {
      authenticators_contains: [
        "0x53d98050a9738da0eac7498d909ea03f6eb03d07fb95877b54ff8acf7712276",
        "0x2b63a8a92b7c67484ab99c4307c7db41b15b9a3c33359cd2b2459fd7f543a9c"
      ]
    }
  ) {
    id
    authenticators
  }
}
```

```gql
{
  spaces(
    where: {
      authenticators_not_contains: [
        "0x2b63a8a92b7c67484ab99c4307c7db41b15b9a3c33359cd2b2459fd7f543a9c"
      ]
    }
  ) {
    id
    authenticators
  }
}
```

They work as you expect them.

PR #321, opened 2024-11-07
fix: read fieldType from currently processed type

Sekhmet:
When using nested where filter we should read field return type from that nested type, not top-level type.

## Test plan

- Same as https://github.com/checkpoint-labs/checkpoint/pull/319
- Plus run following query that was broken:

```gql
{
    proposals(
    where: {
      cancelled: false
      space_in: [
        "0x03ced313394107c43a2f36b63262c00a6422231ac229dc4cedd0d9928d8c3c5f"
      ]
      metadata_: { title_contains_nocase: "" }
    }
  ) {
    id
  }
}
```


bonustrack (approved):
utACK

PR #324, opened 2024-11-08
fix: initialize arrays to string value

Sekhmet:
Arrays are stored as strings in the database so we have to initialize them as such as well.


bonustrack (approved):
utACK

PR #327, opened 2024-12-05
refactor: move startup logic to per-indexer container

Sekhmet:
This PR moves startup logic to container class. This change is internal only, no external API change happened.
Container **will** encapsulate indexer and separate it from main Checkpoint instance.

Right now it moves:
- `reset`/`resetMetadata` calls.
- `validateConfig` call.
- `validateStore` call.
- Helper methods like `getConfigStartBlock` and `getStartBlockNum` as those are used in container.

In the future entire logic for indexer will be encapsulated in Container. Checkpoint will only manage containers calling its life cycle methods (start/reset).

## Test plan
1. Use checkpoint-template.
2. Make sure `resetMetadata` is called before `reset`.
3. Comment out your writer.
4. Start server, you get error message about missing writer.
5. Restore writer.
6. Comment out `resetMetadata` in config.json change Infura node from sepolia to mainnet.
7. Start server, you get error message about network change.
8. Restore original Infura node.
9. Start server, it works as before.

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -7,7 +7,7 @@ import { GqlEntityController } from './graphql/controller';
 import { CheckpointRecord, CheckpointsStore, MetadataId } from './stores/checkpoints';
 import { BaseIndexer, BlockNotFoundError, ReorgDetectedError } from './providers';
 import { createLogger, Logger, LogLevel } from './utils/logger';
-import { getConfigChecksum, getContractsFromConfig } from './utils/checkpoint';
+import { getContractsFromConfig } from './utils/checkpoint';
 import { extendSchema } from './utils/graphql';
 import { createKnex } from './knex';
 import { createPgPool } from './pg';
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
Works well tACK. How a "container" is different from an "indexer"? Also wdyt about exposing "indexer" value in "_metadata" GraphQL entity?

Sekhmet:
> How a "container" is different from an "indexer"?

Indexer just encapsulates logic of how specific network should be processed. If someone wants to create new indexer (for Bitcoin, Solana, etc.) they only need new Indexer (and its Provider).

Container will contain logic of what was previously part of core Checkpoint instance (startup, validation, resetting data, actually running main loop with reorg detection etc.). This part is common across all indexers so we keep it as internal part of Checkpoint.

> Also wdyt about exposing "indexer" value in "_metadata" GraphQL entity?

I had it prepared as a PR, but it included quite a bit of exclusions in the code to only expose it in the metadata entities that I decided to postpone it until regular entities have it too this way it will be supported across all entities and won't require adding special cases in the code.

bonustrack:
Sounds good!

PR #328, opened 2024-12-17
refactor: move lifecycle logic to per-indexer container

Sekhmet:
This PR moves lifecycle to container class. This change is internal only, no external API change happened.
This encapsulates indexer and separates it from main Checkpoint instance.

## Test plan
1. Use checkpoint-template.
2. Make sure `resetMetadata` is called before `reset`.
4. It indexes data as before.


bonustrack (approved):
tACK

PR #329, opened 2025-01-14
feat: add support for multi chain indexing

Sekhmet:
## Summary

Few forced changes for consumers:
- ORM now requires indexer to be specified to know which indexer entity belongs to.
- Things had to move around `CheckpointConfig`, `CheckpointsOptions` and `OverridesConfig` (new, used for database options, like decimals as it needs static config for ORM).

```ts
const mainnetConfig = createConfig('mainnet');
const sepoliaConfig = createConfig('sepolia');

const mainnetIndexer = new starknet.StarknetIndexer(createStarknetWriters('mainnet'));
const sepoliaIndexer = new starknet.StarknetIndexer(createStarknetWriters('sepolia'));

const checkpoint = new Checkpoint(schema, {
  logLevel: LogLevel.Info,
  prettifyLogs: true
});

checkpoint.addIndexer('mainnet', mainnetConfig, mainnetIndexer);
checkpoint.addIndexer('sepolia', sepoliaConfig, sepoliaIndexer);
```

## Test plan 

- Test against updated [checkpoint-template](https://github.com/checkpoint-labs/checkpoint-template/pull/10)
- Test against updated [sx-api](https://github.com/snapshot-labs/sx-monorepo/pull/1092)

### DIFF
--- a/src/bin/index.ts
+++ b/src/bin/index.ts
@@ -6,13 +6,13 @@ import process from 'process';
 import yargs from 'yargs/yargs';
 import { hideBin } from 'yargs/helpers';
 import { codegen } from '../codegen';
-import { CheckpointConfig } from '../types';
+import { OverridesConfig } from '../types';
 
-const DEFAULT_CONFIG_PATH = 'src/config.json';
+const DEFAULT_CONFIG_PATH = 'src/overrides.json';
 const DEFAULT_SCHEMA_PATH = 'src/schema.gql';
 const OUTPUT_DIRECTORY = '.checkpoint';
 
-async function generate(schemaFile: string, configFile: string, format: string) {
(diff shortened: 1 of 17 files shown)


bonustrack (approved):
Works very well! Tried on both APIs. SX-API took a while to start but then it went well.

NIT I've noticed there is now a value `indexer` on each entity entry, wdyt of using the syntax `_indexer`, to make it more obvious this is metadata (not talking about the filter)?

Sekhmet:
Renamed it to `_indexer` both on GraphQL and database level.

bonustrack:
Great! And filter is still like this without underscore right? 
```gql
query spaces (indexer: "abc") { ...
```

Sekhmet:
Yup.

PR #330, opened 2025-01-15
chore: pin starknet

Sekhmet:
Within 5 version starknet makes some changes to exposed types which we depend on. We have to pin it to specific version for now to avoid issues with consumers having unsupported version.


bonustrack (approved):
utACK

PR #335, opened 2025-01-31
chore: replace simple comments with jsdoc

Sekhmet:
This way we get annotations for (some of) our types:

image

image


bonustrack (approved):
Nice, utACK

PR #337, opened 2025-02-03
refactor: reuse instance options for all containers

Sekhmet:
Options should be shared across all containers (they are currently, if you pass options to addIndexer only resetOnChange will be respected, other values are used from main options.


bonustrack (approved):
utACK

PR #341, opened 2025-02-27
fix: on reorg only revert reorganized indexer

Sekhmet:
Previously it would try to revert all entries breaking other indexer's state.

It looks like it was the issue that caused state of one indexer to get nuked (some entries being removed, some ending up as empty) sometimes when running multiple indexers.


bonustrack (approved):
utACK

PR #342, opened 2025-03-05
perf: batch non-derived subentities with loader

Sekhmet:
Previously every non derived 1-to-N mapping would trigger separate database query.

With this change those are batched with dataloader.

For some queries (for example spaces query) it cuts down number of queries from ~20 to 3.

Those queries were made in parallel so speedup is not huge, but makes a difference and should reduce number of spikes in API response times in case of one of those 20 queries being slow.

Before:
image

After:
image


bonustrack (approved):
utACK

PR #343, opened 2025-03-05
fix: do not catch custom Reorg/NoBlock errors

Sekhmet:
Previously we threw and cought our own errors immediatelly, in case of EVM Reorg/NoBlock and in case of Starknet Reorg errors would result in getting block failed... retrying log messages.


bonustrack (approved):
utACK

PR #344, opened 2025-03-06
perf: reduce computations related to event hashes and address comparisons

Sekhmet:
This PR will cache computed event hashes for our sources as well as use faster EVM address comparison (that doesn't add checksum as those are bit expensive).


bonustrack (approved):
utACK

PR #349, opened 2025-03-20
fix: handle errors in latest block checks

Sekhmet:
In case this RPC call fails we need to handle the error. Otherwise it would crash the indexer as it's not handled elsewhere.


bonustrack (approved):
utACK

PR #351, opened 2025-03-21
perf(evm): fetch events for 20 addresses at the time

Sekhmet:
This seems to improve fetching time in some cases by a lot. However it looks like not all nodes handle multiple addresses well enough.

In my testing syncing Sepolia is 6 times faster with Infura API using this approach.


bonustrack (approved):
utACK

PR #353, opened 2025-04-02
fix: sort by block_range by default

Sekhmet:
This change introduces stable ordering for entities - as last step (after user defined orderBy) entries will be ordered by block_range.

This makes it stable as SQL doesn't make any guarantees about order unless ORDER BY is used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/checkpoint",
-  "version": "0.1.0-beta.56",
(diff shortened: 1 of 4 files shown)


bonustrack:
Does this fixes a known issue?

Sekhmet:
Yes, we have (and had for long time) annoying issue caused by order ambiguity that I only noticed recently.
Full PR is incoming.

bonustrack (approved):
utACK

PR #354, opened 2025-04-29
fix: only update latest entry on delete

Sekhmet:
When deleting entry we should only update latest entry, otherwise if entity has multiple versions (it was edited) it will update every one of them, so they will end up having the same (id, indexer, block_range) making it non-unique.


bonustrack (approved):
utACK

PR #356, opened 2025-05-16
refactor: improve time complexity of checking for new sources

Sekhmet:
Our current implementation was O(n^2) which was already bad enough with 150+ sources (20k+ iterations). As it's executed for every transaction in block it ended up with 5M iterations per block.

Before:
image
image

After:
(testing...)

### DIFF
--- a/src/providers/evm/provider.ts
+++ b/src/providers/evm/provider.ts
@@ -226,9 +226,9 @@ export class EvmProvider extends BaseProvider {
       }
 
       const nextSources = this.instance.getCurrentSources(blockNumber);
-      const newSources = nextSources.filter(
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
utACK

Sekhmet:
Updated with new implementation that is faster still. from 19s to 180ms, should be way more CPU efficient too. Will run another test to make sure it still works because I refactored it a bit.

bonustrack:
Awesome, this may help on Arbitrum sync too

Sekhmet:
This will help with Arbitrum too, but it wasn't affected as much as it the issue was bad scaling with number of sources and transactions per block (which are quite low there), but it will prevent scaling issues in the future.

Issue #355, opened 2025-05-15
bug: EVM provider doesn't save checkpoints


Sekhmet:
We are missing call that would store checkpoints for EVM provider.

PR #357, opened 2025-05-19
fix: create checkpoints in EVM provider

Sekhmet:
Previously checkpoints were not created for EVM providers. This adds this missing functionality.

Closes: https://github.com/checkpoint-labs/checkpoint/issues/355

## Test plan

1. Run some API using EVM provider.
2. After a while check for checkpoints.

```gql
{
  _checkpoints(indexer: "sep") {
    id
    indexer
    block_number
  }
}
```


bonustrack (approved):
utACK

PR #364, opened 2025-07-03
feat: improve syncing times

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/workflow/issues/601

This PRs goal is to reduce sync times as much as possible reducing number of RPC requests as much as possible.

Following optimizations are in place:
- EVM preload now filters events by topic so we only process blocks that actually include events we are interested with.
- If events were fetched in preload step they will be used when processing block instead of fetching again.
- `skipBlockFetching` option is added so indexers that don't need it can get even better indexing speed because blocks are not fetched (unless needed post-preload for reorg detection).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/checkpoint",
-  "version": "0.1.0-beta.61",
+  "version": "0.1.0-beta.62",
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

PR #370, opened 2025-07-21
fix: fetch missing logs if new source is added

Sekhmet:
Currently we don't handle a case where we process block/tx and new
events should be handled (due to new source being added).

This PR fixes it by fetching missing logs/events.


bonustrack (approved):
utACK

PR #371, opened 2025-07-21
fix: use blockHash to fetch single block events

Sekhmet:
Closes: https://github.com/snapshot-labs/workflow/issues/613
Depends on: https://github.com/checkpoint-labs/checkpoint/pull/370

Our previous assumption of every block having an event didn't hold: https://sepolia.starkscan.co/block/924016

Instead we use blockHash based event filter which will throw if node doesn't know about given block (as opposed to just returning empty response).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/checkpoint",
-  "version": "0.1.0-beta.62",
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
utACK

PR #372, opened 2025-07-25
fix: only refetch missing events if block comes from preload

Sekhmet:
If event don't come from preload it means we have every event from it, meaning twe don't need to fetch missing ones.

### Test plan

1. Run sx-api with this version starting with latest block on sepolia.
2. Create space on Ethereum Sepolia.
3. Space gets indexed properly (previously it would get stuck there).


bonustrack (approved):
tACK

PR #373, opened 2025-08-19
fix: validate block range hint

Sekhmet:
Closes: https://github.com/snapshot-labs/workflow/issues/639

We had case of RPC using Infura-like range hint, but from/to were either missing or not using hex format which resulted in our code setting itself to NaN.

Now we validate those values are finite.

We also include new Ankr size hint.


bonustrack (approved):
utACK

PR #374, opened 2025-08-19
feat: allow custom pinoOptions

Sekhmet:
This will allow to set custom transports for pino. We will use it to use logtail transport as DigitalOcean truncates logs if too long.


bonustrack (approved):
utACK

PR #375, opened 2025-08-28
chore: bump graphql-parse-resolve-info

Sekhmet:
This version is compatible with GraphQL v16 (it already was before,
but it wasn't listed in peerDependencies).


bonustrack (approved):
utACK, would that fix the graphql issue when using yarn link?

PR #377, opened 2025-08-28
refactor(evm): port EVM provider to viem

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/workflow/issues/611

This PR ports EVM provider from ethers.js v5 to viem.

This brings some quality of live improvements.

#### Client timeout

Now calls made using viem client have 5s timeout. The same timeout is used in `getLogs` request when requesting logs by `blockHash` (single block request).

This should prevent issue of live syncing stopping due to hanging requests which we have experienced.

Timeout is not applied in preload stage as those requests tend to have long response times.

#### Typed events in writers

Now parsed events in writers can be fully typed. This requires constant JSON ABI. To enable this you can type your writers like this:
```typescript
const writer: evm.Writer<typeof MyAbi, 'someEvent'> = ({ event }) => {
  if (!event) return

  event.args.id
  //    ^? - this is now typed using your ABI
}
```

https://github.com/user-attachments/assets/304f09c0-99f5-4c5f-b675-0c23847fb9af

## Test plan

- Link this PR to checkpoint-template [here](https://github.com/snapshot-labs/checkpoint-template/pull/12).
- It still works, evm writers are typed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,10 +20,7 @@
   "prettier": "@snapshot-labs/prettier-config",
   "dependencies": {
     "@ethersproject/abi": "^5.7.0",
-    "@ethersproject/address": "^5.7.0",
-    "@ethersproject/keccak256": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
-    "@ethersproject/strings": "^5.7.0",
     "@graphql-tools/schema": "^8.5.1",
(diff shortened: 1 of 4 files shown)


bonustrack (approved):
tACK

PR #378, opened 2025-08-28
feat: add log message on RPC range hints

Sekhmet:
I've seen some weird behavior. Possibly it could be some provider suggesting range that moves from back into the past, but we don't have logs to verify that.


bonustrack (approved):
utACK

PR #381, opened 2026-04-10
feat: add HyperSync provider for faster event indexing

wa0x6e:
## Summary

Toward https://github.com/snapshot-labs/workflow/issues/787

This PR will add a new HyperSync EVM provider, to use HyperSync (https://envio.dev/#hypersync) as events discovering.

HyperSync is able to return a list of paginated events, for the whole chain, compared to the RPC flow where we had to search for events in the range of 1000 blocks at a time.

This speed up the events searching, down from few weeks to quasi instantaneous. 

The whole remaining indexing time is now due to read/write speed on the database layer (psql).

You can create a free api token, using the free plan. The free plan limits is amply enough to index arbitrum.

## Test plan

In your sx-monorepo app, link your checkpoint dependency to this PR, either by using `bun link`, or editing your package.json to point to your local copy using `file:/` instead of just the version number.

Don't forget to run `yarn build` on checkpoint.

in sx, apps/api/src/evm/index.ts, edit the arb indexer to use

```
const arb1Indexer = new evm.HyperSyncEvmIndexer(createWriters(arb1Config), {
  apiToken: process.env.HYPERSYNC_API_TOKEN || ''
});
```

Add your api token to your apps/api/.env file

Uncomment the arbitrum contracts (keep GMX commented due to issue with timestamp parsing)

Start the sx app with only arbitrum

```
ENABLED_NETWORKS=arb1 bun run dev
```

In the logs, you should see the "preloading" action only once between each events processing, the indexer do not have to find the events by searching blocks by blocks

```
[22:42:10.543] INFO: preloading blocks
    component: "container"
    indexer: "arb1"
    start: 157825417
    end: 451402859
```

Start the sx app with eth

```
ENABLED_NETWORKS=eth bun run dev
```

In the logs, you should see multiple "preloading" events back to back when events are not found

```
[22:58:41.798] INFO: preloading blocks
    component: "container"
    indexer: "eth"
    start: 12006099
    end: 12007099
[22:58:42.516] INFO: preloading blocks
    component: "container"
    indexer: "eth"
    start: 12007100
    end: 12008200
[22:58:43.340] INFO: preloading blocks
    component: "container"
    indexer: "eth"
    start: 12008201
    end: 12009401
``

### CODE REVIEW

File: src/providers/evm/provider.ts [hunk 1]
```diff
@@ -507,15 +523,45 @@ export class EvmProvider extends BaseProvider {
...
     fromBlock: number,
     toBlock: number
   ): Promise<CheckpointRecord[]> {
-    const events = await this.getLogsForSources({
-      fromBlock,
-      toBlock,
-      sources: this.instance.getCurrentSources(toBlock)
-    });
+    const sources = this.instance.getCurrentSources(toBlock);
+    this.blockCache.clear();
```
Sekhmet: Are we clearing the cache each time we call it? This executes multiple times so we will only have cache for only the latest call?

https://github.com/snapshot-labs/checkpoint/blob/7dcf0f02a6e464d9693c9dd68ac4ff04ed27b16a/src/container.ts#L255-L257

Sekhmet (commented):
I think it would be better if we just have two variants of indexers.

`EvmProvider` and `HyperSyncEvmProvider` (and `HyperSyncEvmIndexer` to match). I think it would be better that hypersync config is not part of main config of Checkpoint as providers/indexers are meant to be independent so it can just be passed when doing `new evm.HyperSyncEvmIndexer(writers, { hyperSyncApiKey: "ASDF" })`.

`HyperSyncEvmProvider` could just extend `EvmProvider`. To avoid duplication we could add two methods in `EvmProvider`: `getBlock` and `scanEvents` that `HyperSyncEvmProvider` would just override.

src/providers/evm/provider.ts
wa0x6e: Yes, we're manually clearing it as precaution only, to avoid any leak, since we're already calling `delete` on the cache on read.

Cache should be already clearing by itself after each block, and we're only caching data for the block range, so cache clearing does not have any side effect.

This only needed because we're mixing rpc and hypersync events. Using a 100% hypersync provider, we don't need this cache anymore

wa0x6e:
> I think it would be better if we just have two variants of indexers.
> 
> `EvmProvider` and `HyperSyncEvmProvider` (and `HyperSyncEvmIndexer` to match). I think it would be better that hypersync config is not part of main config of Checkpoint as providers/indexers are meant to be independent so it can just be passed when doing `new evm.HyperSyncEvmIndexer(writers, { hyperSyncApiKey: "ASDF" })`.
> 
> `HyperSyncEvmProvider` could just extend `EvmProvider`. To avoid duplication we could add two methods in `EvmProvider`: `getBlock` and `scanEvents` that `HyperSyncEvmProvider` would just override.

How will you pass the API key from sx apps/api to checkpoint ? Using env var ?

The setup I am thinking of is to have 2 completely independent providers, and let the container.ts decide which one to use (one for live, the other for batch), so hypersync can function 100% without rpc provider.

Sekhmet:
https://github.com/snapshot-labs/sx-monorepo/blob/d5eb6885a6eb531ec8d703fd735c3dbb6aa56db8/apps/api/src/evm/index.ts?plain=1#L70

This would basically become:
```ts
const arb1Indexer = new evm.HyperSyncEvmIndexer(createWriters(arb1Config), { hyperSyncApiKey: process.env.HYPER_SYNC_API_KEY });
```

src/providers/evm/hyper-sync-indexer.ts
```diff
@@ -0,0 +1,44 @@
+import { Logger } from '../../utils/logger';
+import { BaseIndexer, Instance } from '../base';
+import { HyperSyncEvmProvider } from './hyper-sync-provider';
```
bonustrack: Should be a single word for HyperSync filename, its like WalletConnect, we wouldnt do wallet-connect-provider.ts
```suggestion
import { HyperSyncEvmProvider } from './hypersync-provider';
```

wa0x6e: Should we use Hypersync all lowercase too, to match?

bonustrack: No

src/providers/evm/provider.ts
```diff
@@ -549,4 +517,44 @@ export class EvmProvider extends BaseProvider {
     this.log.info('new source added, clearing logs cache');
     this.logsCache.clear();
   }
+
```
bonustrack, line 520: Why we need to change this file and base provider? It would be easier to review if we don't need to changes EvmProvider

wa0x6e: To avoid code duplication, and for proper class inheritance.

Sekhmet (approved):
utACK

It looks good to me, but I haven't tested it myself. If it was tested we can merge it and release it as it shouldn't affect existing indexers as you need to opt-in to use `HyperSyncIndexer`.

wa0x6e:
> utACK
> 
> It looks good to me, but I haven't tested it myself. If it was tested we can merge it and release it as it shouldn't affect existing indexers as you need to opt-in to use `HyperSyncIndexer`.

Should we merge as-is, or include the version bump changes for version release?

Sekhmet:
Let's add version bump.

PR #383, opened 2026-04-16
perf: add partial index for current-version entity lookups

Sekhmet:
Auto-create a partial composite index `(id, _indexer) WHERE upper_inf(block_range)` on every entity table. Speeds up the hot-path query used by Model.loadEntity, save, and delete, and by the GraphQL DataLoader for single-entity resolution, which all filter by `id = ? AND _indexer = ? AND upper_inf(block_range)`. The partial predicate keeps the index small even as historical versions accumulate.

### DIFF
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -273,6 +273,15 @@ export class GqlEntityController {
             )
             .toQuery()
         );
+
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
utACK

PR #384, opened 2026-04-16
perf(evm): cap HyperSync preload range to 1M blocks

Sekhmet:
HyperSync's getPreloadRange returned Infinity, so each preload fetched the full (currentBlock, ~450M) tail in one call. When a new contract source was discovered mid-stream the buffer was discarded and the entire range refetched, taking 10-15 minutes per cycle. Cap the range so each preload covers at most 1M blocks, bounding throwaway work on source discovery while staying large enough to amortize per-request overhead.


bonustrack (approved):
utACK

PR #385, opened 2026-04-16
feat: add state_retention_blocks to purge old historical entity rows

Sekhmet:
Add an opt-in `state_retention_blocks` config option. When set to N, the container periodically deletes closed-range entity rows whose entire validity window falls more than N blocks behind the synced head, and runs that purge once every N synced blocks so the cost stays amortized. Defaults to off; current rows (`upper_inf(block_range)`) are never touched. The retention value must exceed the maximum expected reorg depth or reorg recovery may lose state.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/checkpoint",
-  "version": "0.1.0-beta.69",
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
utACK

PR #390, opened 2026-05-26
feat: add @computed directive and custom resolvers

bonustrack:
## Summary

Adds a `@computed` directive for schema fields resolved at query time via custom SQL subqueries instead of database columns.

- Resolvers are defined with a single `sql` function returning a Knex subquery
- `@computed` fields are skipped from DB table creation and codegen
- `@computed` fields support the full set of `orderBy` and `where` operators based on the underlying type (numeric comparisons, string contains, in/not_in, etc.)
- The SQL subquery is added to SELECT in both list (`queryMulti`) and single-entity (`DataLoader`) queries, so the value is available everywhere via standard GraphQL field resolution

## Usage

Schema:
```graphql
type Space {
  id: String!
  name: String!
  active_proposal_count: Int! @computed
}
```

Checkpoint init:
```typescript
const checkpoint = new Checkpoint(schema, {
  resolvers: {
    Space: {
      active_proposal_count: {
        sql: (knex) =>
          knex('proposals')
            .count('*')
            .where('proposals.space', knex.ref('spaces.id'))
            .where('proposals._indexer', knex.ref('spaces._indexer'))
            .where('proposals.start', '<=', knex.raw("extract(epoch from now())::integer"))
            .where('proposals.max_end', '>', knex.raw("extract(epoch from now())::integer"))
            .where('proposals.cancelled', false)
            .where('proposals.executed', false)
            .where('proposals.vetoed', false)
            .whereRaw('upper_inf(proposals.block_range)')
      }
    }
  }
});
```

Queries:
```graphql
# Sort by active proposals
{ spaces(orderBy: active_proposal_count, orderDirection: desc, first: 10) { id active_proposal_count } }

# Filter spaces with active proposals
{ spaces(where: { active_proposal_count_gt: 0 }, first: 10) { id active_proposal_count } }

# Single-entity query (DataLoader includes the subselect too)
{ space(indexer: "sep", id: "0xabc") { id active_proposal_count } }
```

## Test plan

- [x] `@computed` fields excluded from generated DB tables
- [x] `@computed` fields excluded from codegen models
- [x] `@computed` fields included in orderBy enum and where filters (full type-appropriate set)
- [x] SQL subquery added to SELECT in list queries (queryMulti)
- [x] SQL subquery added to SELECT in single-entity / nested queries (DataLoader)
- [x] ORDER BY works on computed fields (no table prefix)
- [x] WHERE filters work on computed fields via subquery
- [x] Non-computed fields continue to work as before
- [x] Tested locally with sx-monorepo active_proposal_count (list, single, ordered, filtered)

### DIFF
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -26,10 +26,11 @@ import { Knex } from 'knex';
 import pluralize from 'pluralize';
 import { CheckpointsGraphQLObject, MetadataGraphQLObject } from '.';
 import { KnexType } from '../knex';
-import { OverridesConfig } from '../types';
+import { ComputedResolvers, OverridesConfig } from '../types';
 import { getNestedResolver, queryMulti, querySingle } from './resolvers';
 import {
   generateQueryForEntity,
+  getComputedDirective,
   getDerivedFromDirective,
   getNonNullType,
   multiEntityQueryName,
@@ -70,9 +71,14 @@ type WhereResult = {
 export class GqlEntityController {
   private readonly schema: GraphQLSchema;
   private readonly decimalTypes: NonNullable<OverridesConfig['decimal_types']>;
+  private readonly computedResolvers: ComputedResolvers;
   private _schemaObjects?: GraphQLObjectType[];
 
-  constructor(typeDefs: string | Source, config?: OverridesConfig) {
+  constructor(
+    typeDefs: string | Source,
+    config?: OverridesConfig,
+    computedResolvers?: ComputedResolvers
+  ) {
     this.schema = buildSchema(typeDefs);
     this.decimalTypes = config?.decimal_types || {
       Decimal: {
@@ -84,6 +90,7 @@ export class GqlEntityController {
         d: 8
       }
     };
+    this.computedResolvers = computedResolvers || {};
   }
 
   /**
@@ -236,6 +243,8 @@ export class GqlEntityController {
           t.specificType('block_range', 'int8range').notNullable();
 
           this.getTypeFields(type).forEach(field => {
+            if (getComputedDirective(field)) return;
+
             const fieldType =
               field.type instanceof GraphQLNonNull
                 ? field.type.ofType
@@ -381,6 +390,8 @@ export class GqlEntityController {
       };
 
       this.getTypeFields(nestedType).forEach(field => {
+        if (getComputedDirective(field)) return;
+
         // all field types in a where input variable must be optional
         // so we try to extract the non null type here.
         let nonNullFieldType = getNonNullType(field.type);
@@ -639,9 +650,23 @@ export class GqlEntityController {
     const schema = new GraphQLSchema({ query });
 
     if (opts?.addResolvers) {
+      const entityResolvers = this.generateEntityResolvers(entityQueryFields);
+
+      for (const [typeName, fields] of Object.entries(this.computedResolvers)) {
+        if (!entityResolvers[typeName]) entityResolvers[typeName] = {};
+
+        for (const [fieldName, resolver] of Object.entries(fields)) {
+          entityResolvers[typeName][fieldName] = (
+            parent: Record<string, any>,
+            args: Record<string, any>,
+            context: any
+          ) => resolver(parent, args, context);
+        }
+      }
+
       return addResolversToSchema({
         schema,
-        resolvers: this.generateEntityResolvers(entityQueryFields)
+        resolvers: entityResolvers
       });
     }
 
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -392,6 +392,20 @@ export class GqlEntityController {
       this.getTypeFields(nestedType).forEach(field => {
         if (getComputedDirective(field)) {
           orderByValues[field.name] = { value: field.name };
+
+          const fieldType = getNonNullType(field.type);
+          if (fieldType === GraphQLInt) {
+            for (const suffix of ['', '_not', '_gt', '_gte', '_lt', '_lte']) {
+              whereInputConfig.fields[`${field.name}${suffix}`] = {
+                type: GraphQLInt
+              };
+            }
+            for (const suffix of ['_in', '_not_in']) {
+              whereInputConfig.fields[`${field.name}${suffix}`] = {
+                type: new GraphQLList(GraphQLInt)
+              };
+            }
+          }
           return;
         }
 
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -658,15 +658,16 @@ export class GqlEntityController {
       for (const [typeName, fields] of Object.entries(this.computedResolvers)) {
         if (!entityResolvers[typeName]) entityResolvers[typeName] = {};
 
-        for (const [fieldName, resolver] of Object.entries(fields)) {
-          const resolveFn =
-            typeof resolver === 'function' ? resolver : resolver.resolve;
-
+        for (const [fieldName, config] of Object.entries(fields)) {
           entityResolvers[typeName][fieldName] = (
             parent: Record<string, any>,
             args: Record<string, any>,
             context: any
-          ) => resolveFn(parent, args, context);
+          ) => {
+            if (parent[fieldName] !== undefined) return parent[fieldName];
+            if (config.resolve) return config.resolve(parent, args, context);
+            return 0;
+          };
         }
       }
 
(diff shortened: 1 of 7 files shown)


src/graphql/controller.ts
```diff
@@ -381,6 +390,25 @@ export class GqlEntityController {
...
+              whereInputConfig.fields[`${field.name}${suffix}`] = {
+                type: GraphQLInt
+              };
+            }
+            for (const suffix of ['_in', '_not_in']) {
+              whereInputConfig.fields[`${field.name}${suffix}`] = {
+                type: new GraphQLList(GraphQLInt)
+              };
+            }
+          }
```
Sekhmet: So it works only for ints? Non ints can't be used as where condition?

src/graphql/controller.ts
```diff
@@ -639,9 +667,27 @@ export class GqlEntityController {
...
+            args: Record<string, any>,
+            context: any
+          ) => {
+            if (parent[fieldName] !== undefined) return parent[fieldName];
+            if (config.resolve) return config.resolve(parent, args, context);
+            return 0;
+          };
+        }
+      }
+
```
Sekhmet: Not entirely sure how this works? Why are there two different ways for defining resolvers? One is with `resolve` and it seems to only work for single entities and other is `sql` and that one only seems to work for `multi`?

Can't there be just a single way to do this?

src/graphql/controller.ts
bonustrack: Good catch. Refactored to let computed fields fall through the normal field-processing branch, so they pick up the full set of type-appropriate where filters automatically (Int gets gt/lt/in/..., String gets contains/contains_nocase/..., etc.). `buildComputedWhere` in resolvers.ts handles all the suffix variants via subquery.

src/graphql/controller.ts
bonustrack: Agreed, that was confusing. Dropped `resolve` entirely — `sql` is the only way to define a computed field now. The DataLoader (used for single-entity and nested queries) also adds the SQL subquery as a subselect, so the value lands in the row alongside stored columns and resolves via standard GraphQL field resolution. One API, works everywhere.

src/graphql/resolvers.ts
```diff
@@ -97,9 +112,51 @@ export async function queryMulti(
...
+        const placeholders = value.map(() => '?').join(', ');
+        return { sql: `${sub} ${op} (${placeholders})`, bindings: value };
+      }
+      return null;
+    };
+
     Object.entries(where).map((w: [string, any]) => {
-      // TODO: we could generate where as objects { name, column, operator, value }
-      // so we don't have to cut it there
+      for (const [name, config] of Object.entries(entityComputedResolvers)) {
```
Sekhmet, line 149:
```suggestion
      // TODO: we could generate where as objects { name, column, operator, value }
      // so we don't have to cut it there
      for (const [name, config] of Object.entries(entityComputedResolvers)) {
```

src/utils/graphql.ts
```diff
@@ -134,7 +136,21 @@ export const getNonNullType = <T>(type: T): T => {
...
+
+export const getComputedConfigs = (
+  resolvers: ComputedResolvers | undefined,
+  typeName: string
+) =>
+  resolvers
+    ? Object.entries(resolvers).find(
+        ([k]) => k.toLowerCase() === typeName.toLowerCase()
+      )?.[1] || {}
+    : {};
```
Sekhmet, line 156: This expect singular name of type, but sometimes our code passes plural form so this won't work with nested queries I believe:
https://github.com/snapshot-labs/checkpoint/blob/68f85229ff0f73151d5d7df54836974ccb0be184/src/graphql/resolvers.ts?plain=1#L451-L455

I'd suggest just updating how we call `getLoader` to pass entityName instead of (plural) columnName.

```diff
diff --git a/src/graphql/controller.ts b/src/graphql/controller.ts
index cf8beab..4e790d7 100644
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -169,7 +169,7 @@ export class GqlEntityController {
 
             if (itemType instanceof GraphQLObjectType) {
               resolvers[field.name] = getNestedResolver(
-                multiEntityQueryName(itemType)
+                singleEntityQueryName(itemType)
               );
             }
           }
diff --git a/src/graphql/resolvers.ts b/src/graphql/resolvers.ts
index 8fdc56a..895256a 100644
--- a/src/graphql/resolvers.ts
+++ b/src/graphql/resolvers.ts
@@ -394,7 +394,7 @@ export async function querySingle(
 }
 
 export const getNestedResolver =
-  (columnName: string) =>
+  (entityName: string) =>
   async (
     parent: Result,
     args: unknown,
@@ -429,7 +429,7 @@ export const getNestedResolver =
     let result: Record<string, any>[] = [];
     if (!derivedFromDirective) {
       const loaderResult = await getLoader(
-        columnName,
+        entityName,
         'id',
         queryFilter
       ).loadMany(parent[info.fieldName]);
@@ -449,7 +449,7 @@ export const getNestedResolver =
       }
 
       result = await getLoader(
-        columnName,
+        entityName,
         fieldArgument.value.value,
         queryFilter
       ).load(parent.id);

```

src/graphql/resolvers.ts
```diff
@@ -97,9 +112,53 @@ export async function queryMulti(
...
+        const built = buildComputedWhere(
+          `(${config.sql(knex).toQuery()})`,
+          suffix,
+          w[1]
+        );
+        if (built) {
+          query = query.whereRaw(built.sql, built.bindings);
+          return;
+        }
+      }
```
Sekhmet, line 162: All of this is a bit fragile. We are converting it to string and if in custom resolver someones uses Postgres operators that use \\? it will lose escapement and it will break it.

Instead of stringifying query naively we could build query properly and apply it, like this (generated by Claude Code):
```diff
diff --git a/src/graphql/resolvers.ts b/src/graphql/resolvers.ts
index 8fdc56a..bec3434 100644
--- a/src/graphql/resolvers.ts
+++ b/src/graphql/resolvers.ts
@@ -121,25 +121,25 @@ export async function queryMulti(
       _lte: '<='
     };
     const likeOps: Record<string, string> = {
-      _contains: 'LIKE',
-      _not_contains: 'NOT LIKE',
-      _contains_nocase: 'ILIKE',
-      _not_contains_nocase: 'NOT ILIKE'
+      _contains: 'like',
+      _not_contains: 'not like',
+      _contains_nocase: 'ilike',
+      _not_contains_nocase: 'not ilike'
     };
-    const buildComputedWhere = (
-      sub: string,
+    const applyComputedWhere = (
+      currentQuery: Knex.QueryBuilder,
+      sub: Knex.QueryBuilder,
       suffix: string,
       value: any
-    ): { sql: string; bindings: any[] } | null => {
+    ): Knex.QueryBuilder | null => {
+      const expr = knex.raw('(?)', [sub]);
       if (suffix in cmpOps)
-        return { sql: `${sub} ${cmpOps[suffix]} ?`, bindings: [value] };
+        return currentQuery.where(expr, cmpOps[suffix], value);
       if (suffix in likeOps)
-        return { sql: `${sub} ${likeOps[suffix]} ?`, bindings: [`%${value}%`] };
-      if (suffix === '_in' || suffix === '_not_in') {
-        const op = suffix === '_in' ? 'IN' : 'NOT IN';
-        const placeholders = value.map(() => '?').join(', ');
-        return { sql: `${sub} ${op} (${placeholders})`, bindings: value };
-      }
+        return currentQuery.where(expr, likeOps[suffix], `%${value}%`);
+      if (suffix === '_in') return currentQuery.where(expr, 'in', value);
+      if (suffix === '_not_in')
+        return currentQuery.where(expr, 'not in', value);
       return null;
     };
 
@@ -149,13 +149,14 @@ export async function queryMulti(
       for (const [name, config] of Object.entries(entityComputedResolvers)) {
         if (!w[0].startsWith(name)) continue;
         const suffix = w[0].slice(name.length);
-        const built = buildComputedWhere(
-          `(${config.sql(knex).toQuery()})`,
+        const applied = applyComputedWhere(
+          query,
+          config.sql(knex),
           suffix,
           w[1]
         );
-        if (built) {
-          query = query.whereRaw(built.sql, built.bindings);
+        if (applied) {
+          query = applied;
           return;
         }
       }

```

This also makes the code easier to understand.

Sekhmet (approved):
utACK

PR #391, opened 2026-06-11
fix(providers): timeout all provider fetches so hung sockets cannot freeze an indexer

tony8713:
## Symptoms

sx-api indexers silently freeze. On 2026-06-11 we saw the "double box" freeze with chains dying silently around 06:21 and 06:47 UTC: an indexer stops advancing, emits no error, and never recovers until restarted.

## Mechanism

A provider network fetch hangs on a dead/half-open socket. Because there is no timeout, the awaited promise never settles, so the indexer's processing loop is parked forever. The outer `getLogs`/`getEvents` retry loops only re-loop on a thrown error, so a hang (no throw) is invisible and unrecoverable, with nothing in the logs.

Specifically (diagnosed against the installed `@snapshot-labs/checkpoint@0.1.0-beta.70`):
- EVM `_getLogs`: the abort `signal` was set ONLY on the `blockHash` path. The `fromBlock`/`toBlock` range path passed `signal = undefined` to `fetch`, so the range request could hang forever.
- Starknet `getBlockWithReceipts`: a bare `fetch` with no `signal` at all. The `RpcProvider` calls (`getChainId`, `getBlockNumber`, `getBlock`, `getBlockWithTxHashes`, `getEvents`) also have no timeout (starknet 5.x exposes no per-request AbortSignal).
- The retry loops `while (true) { try { ... } catch { sleep(5000) } }` log nothing useful and never trip when the call hangs instead of throwing.

## What changed

- EVM `_getLogs`: always set `AbortSignal.timeout(CLIENT_TIMEOUT)` for every request (both the blockHash and the range path), so every fetch can time out.
- Starknet: add a shared `withTimeout(promise, ms, label)` helper in `utils/helpers.ts` and wrap every `RpcProvider` network call with it; add `AbortSignal.timeout(CLIENT_TIMEOUT)` to the raw `getBlockWithReceipts` fetch. Any hang now becomes a thrown error the existing retry loop already handles.
- Make both retry loops observable: log a `warn` per attempt with the error message and an attempt counter, so a flapping RPC shows up in logs instead of silently.
- Add unit tests for `withTimeout` (resolves before timeout, rejects when the promise never settles).

`CLIENT_TIMEOUT` stays at 5s, matching the existing EVM client timeout.

## Verification

- `yarn build` (tsc): clean
- `yarn lint`: clean
- `yarn test`: 70 passed (incl. 2 new `withTimeout` tests)

## Links

- sx-monorepo #2162
- Diagnosis: sx-monorepo #2162 comment 4681579797

Plain text, no merge requested - leaving for review.

### DIFF
--- a/src/providers/evm/provider.ts
+++ b/src/providers/evm/provider.ts
@@ -368,10 +368,11 @@ export class EvmProvider extends BaseProvider {
       topics?: (string | string[])[];
     } = {};
 
-    let signal: AbortSignal | undefined;
+    // Always set a timeout so a hung socket becomes a thrown error that the
+    // retry loops can handle, instead of silently freezing the indexer forever.
+    const signal = AbortSignal.timeout(CLIENT_TIMEOUT);
 
     if ('blockHash' in filter) {
-      signal = AbortSignal.timeout(CLIENT_TIMEOUT);
       params.blockHash = filter.blockHash;
     }
 
@@ -432,6 +433,7 @@ export class EvmProvider extends BaseProvider {
 
     let currentFrom = fromBlock;
     let currentTo = Math.min(toBlock, currentFrom + MAX_BLOCKS_PER_REQUEST);
+    let attempt = 0;
     while (true) {
       try {
         const logs = await this._getLogs({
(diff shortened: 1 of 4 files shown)


Sekhmet:
Freezes are not HTTP related, but DB connection related.

PR #392, opened 2026-06-19
fix(evm): deduplicate topics

Sekhmet:
This list was being set for each source without deduplication so it could grow by a lot if there is lots of sources.

Generally it shouldn't matter, but:
1. If it grows to huge sizes it will impact requests/server involved.
2. Some providers have limit on the numbers of topics and they might look at the list as-is.


bonustrack (approved):
utACK

PR #393, opened 2026-06-23
fix: handle failures in getNextCheckpointBlock

Sekhmet:
Previously getNextCheckpointBlock failures would kill indexer entirely.

Now if that happens it will retry.


bonustrack (approved):
utACK

PR #394, opened 2026-06-25
fix(hypersync): add rate limitter for HyperSync

Sekhmet:
HyperSync imposes 30 rpm limit and it's easy to hit that limit. If that happens we most likely end up stuck.

This adds rate limiter module that will try to stay under rate limits. If rate limits hit anyway it will use hint from the server to pause requests for given time.


bonustrack (approved):
utACK

PR #395, opened 2026-06-26
chore(evm): handle Edge RPC range errors

Sekhmet:
We are seeing Base indexer getting stuck due to Edge RPC range errors.


bonustrack (approved):
utACK

PR #396, opened 2026-06-30
feat: add syncVersion helper

Sekhmet:
Currently our APIs handle versionTag based reset manually. This is repetetive so it can be included in Checkpoint itself. This is flexible as consumers build the versionTag themselves.

```ts
import Checkpoint from '@snapshot-labs/checkpoint';

const GIT_COMMIT = process.env.GIT_COMMIT;
const ENABLE_SNAPSHOT_X = process.env.ENABLE_SNAPSHOT_X === 'true';
const ENABLE_GOVERNOR_BRAVO = process.env.ENABLE_GOVERNOR_BRAVO === 'true';

async function startIndexer(checkpoint: Checkpoint) {
  addStarknetIndexers(checkpoint);
  addEvmIndexers(checkpoint);

  // undefined when GIT_COMMIT is missing → always resets (safe default)
  const versionTag = GIT_COMMIT
    ? `commit:${GIT_COMMIT}|snapshotX:${ENABLE_SNAPSHOT_X}|governorBravo:${ENABLE_GOVERNOR_BRAVO}`
    : undefined;

  await checkpoint.syncVersion(versionTag);

  checkpoint.start();
}
```

### DIFF
--- a/src/checkpoint.ts
+++ b/src/checkpoint.ts
@@ -8,7 +8,12 @@ import { createPgPool } from './pg';
 import { BaseIndexer } from './providers';
 import { register } from './register';
 import { checkpointConfigSchema } from './schemas';
-import { CheckpointsStore } from './stores/checkpoints';
+import {
+  CheckpointsStore,
+  GLOBAL_INDEXER,
+  METADATA_VALUE_MAX_LENGTH,
+  MetadataId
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
utACK

PR #397, opened 2026-07-11
fix: expose enum fields as String in where inputs

bonustrack:
## Summary

When an entity field is typed as a GraphQL enum, the generated `where` input fields (`field`, `field_not`, `field_in`, `field_not_in`) are now typed as `String` instead of the enum. The output field keeps its enum type (validation and introspection intact).

## Why

Consumers that codegen their types from the query schema (e.g. the sx-monorepo UI) construct filter objects in existing code. If tightening an entity field from `String` to an enum also tightens the filter input, every downstream client is forced to adapt its filter code in lockstep with the schema change. Typing filters as `String` keeps filter shapes stable (and matches hub API filter semantics) while the response side still gets the fully typed enum.

Concrete motivation: adding `state: ProposalState! @computed` to the sx-monorepo API (snapshot-labs/sx-monorepo#2226) without breaking the UI's regenerated `Proposal_filter` types.

## Test plan

- New unit test snapshot: entity with an enum field generates `state: ProposalState!` on the type and `state: String` / `state_not` / `state_in` / `state_not_in` in the filter input
- Full jest suite passes (23 snapshots)
- Verified end-to-end in sx-monorepo: enum output field, string-valued filters, orderBy on the computed enum field, against a live Postgres

### DIFF
--- a/src/graphql/controller.ts
+++ b/src/graphql/controller.ts
@@ -483,15 +483,20 @@ export class GqlEntityController {
         }
 
         if ((nonNullFieldType as GraphQLScalarType).name !== 'Text') {
-          whereInputConfig.fields[`${field.name}`] = { type: nonNullFieldType };
+          const filterType =
+            nonNullFieldType instanceof GraphQLEnumType
+              ? GraphQLString
+              : nonNullFieldType;
+
+          whereInputConfig.fields[`${field.name}`] = { type: filterType };
           whereInputConfig.fields[`${field.name}_not`] = {
-            type: nonNullFieldType
+            type: filterType
           };
           whereInputConfig.fields[`${field.name}_in`] = {
-            type: new GraphQLList(nonNullFieldType)
+            type: new GraphQLList(filterType)
           };
           whereInputConfig.fields[`${field.name}_not_in`] = {
-            type: new GraphQLList(nonNullFieldType)
+            type: new GraphQLList(filterType)
           };
         }
 
--- a/test/unit/graphql/__snapshots__/controller.test.ts.snap
+++ b/test/unit/graphql/__snapshots__/controller.test.ts.snap
@@ -17,6 +17,49 @@ create index \`votes_decimal_index\` on \`votes\` (\`decimal\`);
 create index \`votes_big_decimal_index\` on \`votes\` (\`big_decimal\`)"
 `;
 
+exports[`GqlEntityController generateQueryFields should expose enum fields as String in where inputs 1`] = `
+"type Query {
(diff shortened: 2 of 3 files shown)


Sekhmet (changes requested):
Not sure I understand why we want to change it?

With enums we will get correct typing for inputs: 

Image

With string we just make it accept anything. I'd imagine that if someone generates their client side query types based on schema they expect it to give them benefit of good types, not as loose as possible.

Maybe there is some benefit to this (I can't find a real benefit from PR description), but it's definitely not a "fix".

I see some reference to ProposalState change than forces us to do it, but I don't see why we don't just do that, clean and type-safe:
```diff
diff --git a/apps/ui/src/networks/common/graphqlApi/index.ts b/apps/ui/src/networks/common/graphqlApi/index.ts
index 9545c65bd..e0c26991c 100644
--- a/apps/ui/src/networks/common/graphqlApi/index.ts
+++ b/apps/ui/src/networks/common/graphqlApi/index.ts
@@ -643,30 +643,15 @@ export function createApi(
       filters?: ProposalsFilter,
       searchQuery = ''
     ): Promise<Proposal[]> => {
-      const _filters: ProposalsFilter = clone(filters || {});
+      const { state, labels, ...restFilters } = filters || {};
 
       const metadataFilters: Record<string, any> = {};
       if (searchQuery) metadataFilters.title_contains_nocase = searchQuery;
 
-      const state = _filters.state;
-
-      if (state === 'active') {
-        _filters.start_lte = current;
-        _filters.max_end_gte = current;
-      } else if (state === 'pending') {
-        _filters.start_gt = current;
-      } else if (state === 'closed') {
-        _filters.max_end_lt = current;
+      if (labels?.length) {
+        metadataFilters.labels_contains = labels;
       }
 
-      delete _filters.state;
-
-      if (_filters.labels?.length) {
-        metadataFilters.labels_contains = _filters.labels;
-      }
-
-      delete _filters.labels;
-
       const { data } = await apollo.query({
         query: PROPOSALS_QUERY,
         variables: {
@@ -675,10 +660,11 @@ export function createApi(
           where: {
             space_in: spaceIds,
             cancelled: false,
+            state: state !== 'any' ? state : undefined,
             metadata_: Object.keys(metadataFilters).length
               ? metadataFilters
               : undefined,
-            ..._filters
+            ...restFilters
           }
         }
       });

```

Sekhmet:
This has been addressed at consumer side: https://github.com/snapshot-labs/sx-monorepo/pull/2226

PR #398, opened 2026-07-28
refactor: buffer database access

Sekhmet:
To improve performance CRUD actions on database are now buffered. This reduces number of database round trips within writers.

Due to the internal structure of our entries (for time-travel queries and reorg handling) this is quite convoluted to keep track of this as we need to keep track of current and old versions of entities.

Entries are periodically flushed to database when reaching maximum number of pending rows or on time interval (to prevent failures from reverting too much work).

Once at the tip database is flushed in each block.

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -1,5 +1,6 @@
 import { Knex } from 'knex';
 import { GqlEntityController } from './graphql/controller';
+import { EntityBuffer } from './orm/entity-buffer';
 import {
   BaseIndexer,
   BlockNotFoundError,
@@ -35,6 +36,12 @@ const CHECK_LATEST_BLOCK_INTERVAL = 50;
 
 const DEFAULT_FETCH_INTERVAL = 2000;
 
+/** Maximum time buffered writes are held before a flush is forced. */
+const FLUSH_INTERVAL = 60 * 1000;
+
+/** Maximum entity buffer size before a flush is forced. */
+const MAX_BUFFER_SIZE = 10000;
+
 export class Container implements Instance {
   private indexerName: string;
 
@@ -51,12 +58,14 @@ export class Container implements Instance {
   private activeTemplates: TemplateSource[] = [];
   private cpBlocksCache: number[] | null = [];
   private blockHashCache: { blockNumber: number; hash: string } | null = null;
+  private entityBuffer: EntityBuffer;
 
   private preloadStep: number = BLOCK_PRELOAD_START_RANGE;
   private preloadedBlocks: number[] = [];
   private preloadEndBlock = 0;
 
   private lastPurgeBlock: number | null = null;
+  private lastFlushAt = Date.now();
 
   constructor(
     indexerName: string,
@@ -70,6 +79,8 @@ export class Container implements Instance {
     opts?: CheckpointOptions
   ) {
     this.indexerName = indexerName;
+    this.entityBuffer = register.getEntityBuffer(indexerName);
+    this.entityBuffer.reset();
     this.log = log.child({ component: 'container', indexer: indexerName });
     this.knex = knex;
     this.store = store;
@@ -126,7 +137,10 @@ export class Container implements Instance {
       return this.blockHashCache.hash;
     }
 
-    return this.store.getBlockHash(this.indexerName, blockNumber);
+    return (
+      this.entityBuffer.getBlockHash(blockNumber) ??
+      this.store.getBlockHash(this.indexerName, blockNumber)
+    );
   }
 
   private addSource(source: ContractSourceConfig) {
@@ -188,19 +202,15 @@ export class Container implements Instance {
   public async setBlockHash(blockNum: number, hash: string) {
     this.blockHashCache = { blockNumber: blockNum, hash };
 
-    return this.store.setBlockHash(this.indexerName, blockNum, hash);
+    this.entityBuffer.setBlockHash(blockNum, hash);
   }
 
   public async setLastIndexedBlock(block: number) {
-    await this.store.setMetadata(
-      this.indexerName,
-      MetadataId.LastIndexedBlock,
-      block
-    );
+    this.entityBuffer.setLastIndexedBlock(block);
   }
 
   public async insertCheckpoints(checkpoints: CheckpointRecord[]) {
-    await this.store.insertCheckpoints(this.indexerName, checkpoints);
+    this.entityBuffer.addCheckpoints(checkpoints);
   }
 
   /**
@@ -351,12 +361,15 @@ export class Container implements Instance {
 
       try {
         register.setCurrentBlock(this.indexerName, BigInt(blockNumber));
+        this.entityBuffer.beginBlock();
 
         const initialSources = this.getCurrentSources(blockNumber);
         const parentHash = await this.getBlockHash(blockNumber - 1);
         const nextBlockNumber = await this.indexer
           .getProvider()
           .processBlock(blockNumber, parentHash);
+        this.entityBuffer.commitBlock();
+
         const sources = this.getCurrentSources(nextBlockNumber);
 
         if (initialSources.length !== sources.length) {
@@ -364,24 +377,18 @@ export class Container implements Instance {
         }
 
         blockNumber = nextBlockNumber;
-
-        if (this.config.state_retention_blocks) {
-          if (this.lastPurgeBlock === null) {
-            this.lastPurgeBlock = blockNumber;
-          } else if (
-            blockNumber - this.lastPurgeBlock >=
-            this.config.state_retention_blocks
-          ) {
-            await this.purgeOldState(blockNumber);
-            this.lastPurgeBlock = blockNumber;
-          }
-        }
       } catch (err) {
+        this.entityBuffer.rollbackBlock();
+
         if (err instanceof BlockNotFoundError) {
           if (this.config.optimistic_indexing) {
             try {
+              this.entityBuffer.beginBlock();
               await this.indexer.getProvider().processPool(blockNumber);
+              this.entityBuffer.commitBlock();
+              await this.flushBuffer(blockNumber);
             } catch (err) {
+              this.entityBuffer.rollbackBlock();
               this.log.error(
                 { blockNumber: blockNumber, err },
                 'error occurred during pool processing'
@@ -409,7 +416,50 @@ export class Container implements Instance {
         }
 
         await sleep(this.config.fetch_interval || DEFAULT_FETCH_INTERVAL);
+        continue;
       }
+
+      await this.maybeFlush(blockNumber);
+    }
+  }
+
+  private async flushBuffer(currentBlock: number) {
+    while (true) {
+      try {
+        await this.entityBuffer.flush();
+        this.lastFlushAt = Date.now();
+
+        break;
+      } catch (err) {
+        this.log.error({ err }, 'flush failed, retrying');
+        await sleep(this.config.fetch_interval || DEFAULT_FETCH_INTERVAL);
+      }
+    }
+
+    if (!this.config.state_retention_blocks) return;
+
+    try {
+      if (this.lastPurgeBlock === null) {
+        this.lastPurgeBlock = currentBlock;
+      } else if (
+        currentBlock - this.lastPurgeBlock >=
+        this.config.state_retention_blocks
+      ) {
+        await this.purgeOldState(currentBlock);
+        this.lastPurgeBlock = currentBlock;
+      }
+    } catch (err) {
+      this.log.error({ err }, 'purging old state failed');
+    }
+  }
+
+  private async maybeFlush(currentBlock: number) {
+    const atTip = currentBlock > this.preloadEndBlock;
+    const stale = Date.now() - this.lastFlushAt >= FLUSH_INTERVAL;
+    const full = this.entityBuffer.size >= MAX_BUFFER_SIZE;
+
+    if (atTip || stale || full) {
+      await this.flushBuffer(currentBlock);
     }
   }
 
@@ -438,6 +488,8 @@ export class Container implements Instance {
   private async handleReorg(blockNumber: number) {
     this.log.info({ blockNumber }, 'handling reorg');
 
+    await this.flushBuffer(blockNumber);
+
     let current = blockNumber - 1;
     let lastGoodBlock: null | number = null;
     while (lastGoodBlock === null) {
@@ -464,7 +516,7 @@ export class Container implements Instance {
       }
     }
 
-    const entities = await this.entityController.schemaObjects;
+    const entities = this.entityController.schemaObjects;
     const tables = entities.map(entity =>
       getTableName(entity.name.toLowerCase())
     );
@@ -490,6 +542,7 @@ export class Container implements Instance {
     // TODO: when we have full transaction support, we should include this in the transaction
     await this.store.removeFutureData(this.indexerName, lastGoodBlock);
 
+    this.entityBuffer.reset();
     this.cpBlocksCache = null;
     this.blockHashCache = null;
 
@@ -499,6 +552,8 @@ export class Container implements Instance {
   }
 
   public async reset() {
+    this.entityBuffer.reset();
+
     await this.store.setMetadata(
       this.indexerName,
       MetadataId.LastIndexedBlock,
--- a/src/container.ts
+++ b/src/container.ts
@@ -42,6 +42,9 @@ const FLUSH_INTERVAL = 60 * 1000;
 /** Maximum entity buffer size before a flush is forced. */
 const MAX_BUFFER_SIZE = 10000;
 
+/** Maximum delay between flush retries when backing off. */
+const MAX_FLUSH_RETRY_DELAY = 60 * 1000;
+
 export class Container implements Instance {
   private indexerName: string;
 
@@ -424,6 +427,7 @@ export class Container implements Instance {
   }
 
   private async flushBuffer(currentBlock: number) {
+    let retryDelay = this.config.fetch_interval || DEFAULT_FETCH_INTERVAL;
     while (true) {
       try {
         await this.entityBuffer.flush();
@@ -431,8 +435,12 @@ export class Container implements Instance {
 
         break;
       } catch (err) {
-        this.log.error({ err }, 'flush failed, retrying');
-        await sleep(this.config.fetch_interval || DEFAULT_FETCH_INTERVAL);
+        this.log.error(
+          { err, currentBlock, bufferSize: this.entityBuffer.size },
+          'flush failed, retrying'
+        );
+        await sleep(retryDelay);
+        retryDelay = Math.min(retryDelay * 2, MAX_FLUSH_RETRY_DELAY);
       }
     }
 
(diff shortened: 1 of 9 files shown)


src/orm/entity-buffer.ts
```diff
@@ -0,0 +1,442 @@
...
+    tableName,
+    id
+  }: {
+    tableName: string;
+    id: string;
+  }): Promise<EntityValues | null> {
+    const key = this.getKey(tableName, id);
+    const entity = this.entities.get(key);
+    if (entity) {
+      if (entity.openVersion) return { ...entity.openVersion.values };
```
tony8713, line 112: `load()` returns whatever the writer last put in the buffer, but once a flush has happened the same call returns the value round-tripped through Postgres. For any column type where those differ, the JS type a writer sees now depends on flush timing.

Verified with a `BigInt!` field (`int8` column), identical writer code, same workload:

```
master                        b3 v1.power typeof=string value=1000
this PR, buffered window      b3 v1.power typeof=number value=1000   <-- changed
this PR, at tip (flush/block) b3 v1.power typeof=string value=1000
```

`node-pg` returns `int8` as a string; the writer had set a `number`. So the type flips depending on whether a flush landed between the write and the read — which in practice means **backfill and tip behave differently**, since `maybeFlush` makes `atTip` flush every block while a backfill buffers thousands. Same for `decimal` columns (numeric comes back as a string) and `bigint` values a writer sets as `0n`.

sx writers mostly coerce (`BigInt(proposal.scores_total) + BigInt(vote.vp)`), so I could not demonstrate a live break, and I'm not blocking on it. But `x + 1n` vs `'1000' + 1n`, or `a > b` on two `BigInt` fields, silently changes meaning, and it'd be a miserable bug to chase because it only reproduces at one of the two speeds.

Cheapest fix is to make the buffer the single source of truth for shape — normalise on the way in (or serialise the version values the same way the DB would) so `load()` is type-stable regardless of flush timing. At minimum worth a comment here saying reads inside a window are un-normalised.

src/container.ts
```diff
@@ -409,7 +416,50 @@ export class Container implements Instance {
...
+
+  private async flushBuffer(currentBlock: number) {
+    while (true) {
+      try {
+        await this.entityBuffer.flush();
+        this.lastFlushAt = Date.now();
+
+        break;
+      } catch (err) {
+        this.log.error({ err }, 'flush failed, retrying');
```
tony8713: No block context on this line, so a wedged indexer gives you no idea which block or entity is poisonous. `master` logged `blockNumber` at the failing write; here the whole window collapses into one message.

What it actually looks like on a permanent failure (this is one line, repeating every `fetch_interval` forever):

```
flush failed, retrying
  err: insert into "votes" (...) values ($1, int8range($2, NULL), $3, $4, $5),
       ($6, ...), ... ($196, int8range($197, NULL), $198, $199, $200)
       - value too long for type character varying(256)
```

That's 200 placeholders for a 40-block window. At `MAX_BUFFER_SIZE` it's ~50k placeholders per line, every 2s, forever, into Better Stack. Suggest `this.log.error({ err, currentBlock, bufferSize: this.entityBuffer.size }, 'flush failed, retrying')` plus some backoff, so the volume is survivable and the message is actionable.

src/container.ts
```diff
@@ -35,6 +36,12 @@ const CHECK_LATEST_BLOCK_INTERVAL = 50;
 
 const DEFAULT_FETCH_INTERVAL = 2000;
 
+/** Maximum time buffered writes are held before a flush is forced. */
+const FLUSH_INTERVAL = 60 * 1000;
```
tony8713, line 40: There's no drain anywhere in the repo — `grep -rn "SIGTERM|SIGINT|beforeExit|process.on" src/` returns nothing — so a deploy or pod rollover discards the whole window. Measured: SIGKILL at block 7000 after a flush at 5000 threw away 2000 blocks of work.

That's *safe* (the cursor goes with it and the blocks replay), but it's a new cost this PR introduces — `master` committed every block, so a restart lost nothing. On our infra pod rollovers are routine, so every deploy now buys a re-index of up to 60s / 10000 buffer units.

A `process.on('SIGTERM')` that calls `flushBuffer()` before exiting, or an exported `Checkpoint.stop()` that consumers can wire up, would make deploys close to free. Worth a follow-up if you'd rather not grow this PR.

src/container.ts
```diff
@@ -409,7 +416,50 @@ export class Container implements Instance {
...
+      }
+    } catch (err) {
+      this.log.error({ err }, 'purging old state failed');
+    }
+  }
+
+  private async maybeFlush(currentBlock: number) {
+    const atTip = currentBlock > this.preloadEndBlock;
+    const stale = Date.now() - this.lastFlushAt >= FLUSH_INTERVAL;
+    const full = this.entityBuffer.size >= MAX_BUFFER_SIZE;
```
tony8713, line 467: Two small things on this line.

**The bound is row count, not bytes.** `size` counts entities + versions + checkpoints + block hashes, so peak memory scales with average entity payload with no cap. sx uses `Text` for `body` / `about` / `metadata`. Measured with a 10KB payload per entity:

```
block=300  bufferSize=1800   heapUsed=21.2MB  rss=168MB
block=900  bufferSize=5400   heapUsed=53.6MB  rss=208MB
block=1500 bufferSize=9000   heapUsed=82.8MB  rss=257MB
>>> MAX_BUFFER_SIZE reached at block 1667: bufferSize=10002 heapDelta=96.3MB rss=284MB
```

~100MB at the cap for a 10KB average, and it's linear — a schema with fatter `Text` fields scales straight up from there. Worth either a byte-aware cap or a comment recording the assumption.

**`this.entityBuffer.size` is O(entities) and runs once per block.** It iterates the whole map to sum `closedVersions.length`. Part of why sustained throughput drops from 8333 to 1481 blocks/sec as the buffer fills. Maintaining the count incrementally in `save`/`delete`/`addCheckpoints` would be cheap. Still ~50x faster than `master` either way, so this is purely a nit.

src/orm/entity-buffer.ts
```diff
@@ -0,0 +1,442 @@
...
+  async flush() {
+    const knex = this.getKnex();
+    const statements = this.prepareFlush(knex);
+
+    if (statements.length === 0) {
+      this.reset();
+      return;
+    }
+
+    await knex.transaction(async trx => {
```
tony8713, line 393: Nit: `flush()` is the only part of this file the tests don't reach.

```
src/orm/entity-buffer.ts | 95.83 | 93.88 | 381-398
```

The unit tests are genuinely good on the in-memory semantics and on `prepareFlush` SQL generation — but the transaction itself, which is the piece carrying the durability guarantee, is untested. I exercised it by hand against real Postgres and it's correct (statement ordering, atomicity of data + cursor, retry-after-failure, empty-buffer no-op), so this is a suggestion rather than a request: one integration test that flushes into a real DB and asserts the resulting `block_range` rows would lock in the property that matters most.

tony8713 (approved):
LGTM. Approving.

Given this is the indexer behind Snapshot X, I didn't want to review the SQL by reading it, so I stood the branch up against a local Postgres 16 and drove a real `Container` through scripted workloads — including killing the process mid-buffer and injecting flush failures. **Everything on the durability list checks out, and the branch fixes two real bugs that exist on `master` today.** All findings below are non-blocking.

---

### The important part: it cannot advance the cursor past lost writes

This was my main worry — buffering is exactly the change that can produce "cursor moved, rows never landed, no error". It can't here, for two structural reasons, and I confirmed both:

1. `last_indexed_block` is buffered like everything else and emitted as the last statement of the same `knex.transaction` in `flush()`, so data and cursor commit or roll back together.
2. `flushBuffer()` retries forever and never advances on failure.

**Permanent flush failure** (a writer emits a 400-char value for a `varchar(256)` column at block 20, so every attempt fails):

```
[flush error]  ... - value too long for type character varying(256)
[t+4s]  votes=0 max_count=-1 last_indexed_block=0 _blocks=0 flushErrors=2
[t+8s]  votes=0 max_count=-1 last_indexed_block=0 _blocks=0 flushErrors=4
[FINAL] votes=0 max_count=-1 last_indexed_block=0 _blocks=0 flushErrors=8
```

Error surfaced, cursor pinned at 0. Correct — it wedges loudly instead of skipping blocks.

**Transient flush failure** (a `BEFORE INSERT` trigger raises once, then is dropped):

```
[harness] flush error #1 / #2: sabotage: injected flush failure
[harness] sabotage trigger removed
[FINAL] votes=3126 last_indexed_block=3127 flushErrors=2

total_rows | distinct_ids | mn |  mx
      3128 |         3128 |  1 | 3128
duplicates: 0 rows
```

The buffer survives a failed flush intact and replays it exactly once — no gaps, no duplicates.

**Crash / SIGTERM / SIGKILL / unhandled rejection**, fired at block 25 with an un-flushed buffer — all four modes:

```
votes=0   last_indexed_block=0   blocks=0
```

Writes are lost, but the cursor is lost with them, so restart replays. Verified end to end: SIGKILL at block 7000 after a real flush at 5000 left `votes=5000, last_indexed_block=5000`; restarting and running on gave

```
total=10000  distinct_ids=10000  missing_in_1_10000=0  dupes=0
overlapping_ranges=0  last_indexed_block=10000
```

Replay is complete and idempotent.

### Ordering

Correct. `prepareFlush` emits every `close` before every `insert`, and the GiST exclusion constraint on `(id, _indexer, block_range)` would reject any mistake, so this is enforced rather than assumed. I ran a window mixing rows that already existed in the DB with rows created inside the window, plus deletes and a re-create:

```
e1 count=1 range=[1,6)    e1 count=2 range=[6,8)    e1 count=3 range=[8,)
e2 count=1 range=[1,9)    e2 count=9 range=[12,)          <- deleted b9, recreated b12
e3 count=1 range=[1,10)                                   <- updated then deleted in b10
e4 count=1 range=[7,8)    e4 count=2 range=[8,11)         <- created b7, deleted b11
overlapping_ranges=0
```

I also ran the same 40-block workload three ways — `master`, this branch fully buffered (one flush at the end), and this branch at tip (flush every block) — and the non-empty rows are byte-identical across all three.

### Two bugs this fixes on the way

**1. Two `save()` calls on one `Model` instance in one block wedges `master` permanently.** `master`'s `save()` never sets `exists`, so the second call is another `_insert()`:

```
error: conflicting key value violates exclusion constraint "votes_id__indexer_block_range_excl"
detail: Key (id, _indexer, block_range)=(v5, eth, [8,)) conflicts with existing key (v5, eth, [8,))
code: 23P01
```

...retried forever, indexer stuck. This branch coalesces it into a single version and produces `v5 count=99 range=[8,)`. Nice catch.

**2. `master` writes EMPTY-range junk rows.** Any same-block create+delete, update+delete, or create+update makes `_update`/`_delete` close a range to `int8range(N, N)`. In the 40-block workload `master` left 4 dead rows; this branch leaves 0.

Also `_blocks` is now an upsert instead of `master`'s bare insert, which is the right call for replay. And the `seenPoolTransactions` move in the Starknet provider is correct and complete — EVM has no `processPool`, so there's no matching site to fix.

### Dedup / upsert

Checked against real PG rather than assumed, since the two branches differ:

```sql
INSERT ... VALUES ('a','eth',1), ('a','eth',1) ON CONFLICT DO NOTHING;   -- INSERT 0 2, fine
INSERT ... VALUES ('a','eth',1), ('a','eth',9) ON CONFLICT DO UPDATE ...;
ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time
```

Checkpoints use `.ignore()` so in-statement duplicates are safe; block hashes use `.merge()` and are collapsed through `latestHashes` first, which is exactly what's needed. Empty-buffer flush is a clean no-op — it returns before opening a transaction.

### Throughput

It delivers. Same workload (one insert + one read-modify-write per block), 3000 blocks:

```
master   99.21s    30 blocks/sec
this PR   0.36s  8333 blocks/sec
```

and 20000 blocks including four real flushes still holds **1481 blocks/sec**, ~50x.

---

Non-blocking comments inline. The one I'd most like you to look at is the read-path type drift on `load()`.

src/container.ts
Sekhmet: SIGTERM might not be ideal because it could flush in middle of processing so it would result in broken state. For now we can accept it can lose some process on restarts.

src/orm/entity-buffer.ts
tony8713: Fixed by `e72550e`. Confirmed by measurement rather than by reading — same schema, same writer, all three states now agree:

```
never-set (constructor default)    typeof=bigint  value=0
after write, BUFFERED (no flush)   typeof=bigint  value=1000
after write, FLUSHED (from pg)     typeof=bigint  value=1000

raw pg column value:               typeof=string  value="1000"   <- what the getter absorbs
```

And the case that silently corrupted before: `power = power + 1n` across two flushes now gives `1002`. Mixing a number throws (`TypeError: Invalid mix of BigInt and other type in addition`) instead of string-concatenating. Resolving.

---

One thing worth writing down before this thread goes. **Not a blocker, nothing to do in this PR.**

The same buffered-vs-tip drift still exists for `Float` and for `Decimal` at non-zero scale. It's a *value* drift rather than a type drift, which makes it quieter than the BigInt one — `typeof` is stable in both states and nothing throws:

| field | buffered | flushed | column |
|---|---|---|---|
| `Float` | `number` `0.123456789` | `number` `0.12345679` | `float4` |
| `Decimal` | `string` `"1.5"` | `string` `"1.50"` | `numeric(10,2)` |
| `BigDecimal` | `string` `"1.5"` | `string` `"1.50000000"` | `numeric(20,8)` |

(`Float` lands as `float4` because knex's pg dialect hardcodes `ColumnCompiler_PG.prototype.floating = 'real'`, so the `options: [23]` at `controller.ts:574` is discarded.)

Not currently exercised by any sx writer, which is why I'm recording it rather than raising it: all five `Float!` fields in `apps/api` (`scores_*_parsed`, `vp_parsed`) are write-only, and the decimal fields that *are* read back are configured at scale 0 — `BigDecimalVP {p:60,d:0}` in `apps/api`, `BigIntVP {p:80,d:0}` in `delegates-api` — so they round-trip byte-identically today.

The reason it's worth a note: `apps/api` does `BigInt(proposal.scores_total)` at `snapshot-x/writers.ts:1048` and eleven other sites, and that is safe *only* because `d: 0`. Set a non-zero scale in `overrides.json` and those calls become a hard `SyntaxError: Cannot convert 1500.50000000000000000000 to a BigInt` — in exactly one of the two flush states, so it would reproduce at one indexing speed and not the other.

src/container.ts
tony8713: Agreed, and the concern is right — a naive handler could flush a half-processed block.

Worth noting for whenever this comes back: it doesn't have to be mid-block. `maybeFlush` at `container.ts:422` is already the last statement in the `while (true)` body, after `commitBlock()` at `:371`, so a flag checked at that point drains at a clean block boundary — no partial state, no rollback needed. And the journal (`beginBlock` / `rollbackBlock`) covers the case where you do get interrupted mid-block.

Happy to leave it as-is. Resolving.

src/container.ts
tony8713: Re-measured at `e72550e` (`src/container.ts` is byte-identical to `408e5c1`, so this still stands) — but **I overstated the volume above and want to correct it.**

There is no ~50k-placeholder line. `prepareFlush` chunks at `MAX_INSERT_ROWS = 1000`, so the failing statement is capped at 1000 rows and the log line plateaus:

```
blocks=40     bufferSize=200    err.message=11226 chars  placeholders=1200
blocks=200    bufferSize=1000   err.message=60026 chars  placeholders=6000
blocks=1000   bufferSize=5000   err.message=60026 chars  placeholders=6000
blocks=2000   bufferSize=10000  err.message=60026 chars  placeholders=6000
```

So the real shape is a **bounded ~60KB line, repeated every `fetch_interval` forever** — ~1.7MB/min at the 2000ms default. Bounded per line, unbounded in total.

The part I'd still ask for is the context, not the volume: there's no block number and no buffer size on that line, so a wedged indexer tells you nothing about *which* block is poisonous, and the retry at `container.ts:427-437` is a fixed-interval `while (true)` with no backoff.

Same suggestion as before: `this.log.error({ err, currentBlock, bufferSize: this.entityBuffer.size }, 'flush failed, retrying')`.

src/orm/entity-buffer.ts
```diff
@@ -0,0 +1,442 @@
...
+  reset() {
+    this.entities.clear();
+    this.checkpoints = [];
+    this.blockHashes = [];
+    this.lastIndexedBlock = null;
+    this.commitBlock();
+  }
+
+  private getKey(tableName: string, id: string): string {
+    return `${tableName} ${id}`;
```
tony8713: There's a literal NUL byte (`0x00`) in this template string — it renders as a space in the diff above, but the separator is a raw control byte rather than an escape.

The *choice* is right, to be clear: a space would collide (`getKey("a", "b c")` and `getKey("a b", "c")` both give `"a b c"`), NUL can't. It's the encoding I'd change, because a raw NUL makes the file read as binary:

```
$ file src/orm/entity-buffer.ts
src/orm/entity-buffer.ts: data

$ grep -n "getKey" src/orm/entity-buffer.ts
$ echo $?
1
```

That's a silent zero-match — no output, no `Binary file matches`. ripgrep prints `binary file matches` and skips the lines. It's the only file in `src/` that does this. I hit it myself on this pass: my `grep -rn "SIGTERM\|process\.on" src/` sweep was quietly skipping this file until I added `-a`.

Git doesn't flag it either — the byte is at offset 10954, past git's 8000-byte binary-sniff window, so `--numstat` reports `442 0` and the diff renders as ordinary text.

`\0` or `\u0000` gives the identical key with none of that. Minor related thing: `prepareFlush` at `:265` builds its own composite key with a plain space, so the file currently has two separator conventions.

tony8713 (approved):
Re-reviewed at `e72550e`. Still an approve.

**Resolved**
- `entity-buffer.ts:112` read-path type drift — fixed, confirmed by measurement (all three states now return a true `bigint`, and `+=` across a flush persists 1002 instead of corrupting). Details in the thread, plus one `Float`/`Decimal` note recorded there for later.
- `container.ts:40` shutdown drain — your call, agreed.

**Still open.** `git diff 408e5c1 e72550e -- src/container.ts src/orm/` is empty, so these three are byte-identical to what I reviewed and stand as written:

- `container.ts:434` flush-error log — **I overstated this and have corrected it in the thread.** `prepareFlush` chunks at `MAX_INSERT_ROWS = 1000`, so it plateaus at a bounded ~60KB line, not the ~50k placeholders I claimed. The ask is now just the missing block context, not the volume.
- `container.ts:459` — re-measured on head. Same `bufferSize=10002` in every run, heap at the cap set entirely by payload size: **14.3 / 43.6 / 102.2 / 199.9 MB** for 1 / 4 / 10 / 20 KB entities. `size` getter: 4.9µs at 1k entities → 104µs at 10k → 516µs at 40k.
- `entity-buffer.ts:393` — `flush()` is still the only uncovered code in the file: `95.83 | 93.88 | 381-398`, unchanged. It's also the only caller path that opens the transaction, and `container.ts` isn't imported by any test.

**New** — one inline comment below, on a stray NUL byte in `entity-buffer.ts`. Cosmetic in effect, but it makes the file invisible to `grep`, so it's worth the one-character fix.

Tests on head: **94 pass / 0 fail**, `tsc` clean.

bonustrack (approved):
utACK

PR #400, opened 2026-08-03
fix: avoid unrecoverable database write failures

Sekhmet:
SX API in some cases writes bad data into ORM. This previously was handled as failures would be surfaced on `.save` call.
But now as writes are buffered this becomes an issue when writes are flushed and at this point it's not possible to handle that at writer level.

To avoid that we add two edge case handling in ORM setter:
1. `varchar` overflows 256 characters.
2. `varchar` or `string` based value contains NUL byte.

When writer tries to set such value it will throw in-place where writer can react to it.

There was another issue with `jsonb` index size. This index isn't actually used for the queries we use (not applicable for operators we use), so we simply remove it.

### DIFF
--- a/src/codegen.ts
+++ b/src/codegen.ts
@@ -107,6 +107,28 @@ export const getJSType = (
   return { isNullable, isList, baseType };
 };
 
+/** Maximum length of varchar columns created for String/ID fields. */
+const VARCHAR_LENGTH = 256;
+
+const isVarcharField = (type: GraphQLType): boolean => {
+  const nonNullType = type instanceof GraphQLNonNull ? type.ofType : type;
+
+  if (nonNullType === GraphQLString || nonNullType === GraphQLID) return true;
+
+  if (nonNullType instanceof GraphQLObjectType) {
+    const idField = nonNullType.getFields()['id'];
+
+    return (
+      idField !== undefined &&
+      idField.type instanceof GraphQLNonNull &&
+      idField.type.ofType instanceof GraphQLScalarType &&
+      ['String', 'ID'].includes(idField.type.ofType.name)
+    );
+  }
+
+  return false;
+};
+
 const isPersistedField = (field: GraphQLField<any, any>) => {
   if (getComputedDirective(field)) return false;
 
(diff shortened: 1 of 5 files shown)


bonustrack (approved):
utACK

src/graphql/controller.ts
```diff
@@ -265,7 +265,7 @@ export class GqlEntityController {
               column = column.notNullable();
             }
 
-            if (!['text', 'json'].includes(sqlType.name)) {
+            if (!['text', 'json', 'jsonb'].includes(sqlType.name)) {
```
tony8713, line 268: Worth saying in the PR for whoever deploys: `createEntityStores` is drop-and-recreate and only runs from `reset()`, so bumping the package won't remove the btree indexes that already exist on live databases. Those keep failing writes until a reindex. `DROP INDEX <table>_<column>_index;` unblocks an affected one in place. Not a change request.

src/codegen.ts
```diff
@@ -223,6 +255,24 @@ export const codegen = (
...
+      if (baseType === 'string' || baseType === 'string[]') {
+        const nulCheck = isList
+          ? `value.some(item => typeof item === 'string' && item.includes('\\u0000'))`
+          : `value.includes('\\u0000')`;
+        contents += isNullable
+          ? `    if (value !== null && ${nulCheck}) {\n`
+          : `    if (${nulCheck}) {\n`;
+        contents += `      throw new Error('Value for ${modelName}.${field.name} contains a NUL character');\n`;
+        contents += `    }\n\n`;
+      }
```
tony8713, line 270: `value !== null` is the wrong check here: it lets every other non-string through into `[...value]` and `value.includes(...)`, which throws an opaque `TypeError` on values that write fine today. A number into a `Decimal` field or an Int-keyed ref, and `undefined` into a nullable String, all work on master and throw on this branch.

`typeof value === 'string' && ...` on both checks fixes it, and the `isNullable` branching then collapses since that already excludes null and undefined.

test/unit/codegen.test.ts
```diff
@@ -202,3 +203,120 @@ describe('codegen', () => {
...
   });
 });
+
+describe('generated models', () => {
+  const extendedSchema = extendSchema(SCHEMA_SOURCE);
+  const controller = new GqlEntityController(extendedSchema);
+
+  const source = codegen(controller, {}, 'javascript')
+    .replace(/^import .*\n\n/, '')
+    .replace(/export /g, '');
```
tony8713, line 213: Two gaps: no case for non-string setter input (the `TypeError` I flagged in `codegen.ts`), and nothing covers the jsonb change at all. The existing `createEntityStores` snapshot is an easy place to assert that no btree index is created on a list column.

tony8713 (approved):
Ran both halves against a local Postgres 16. They do what the description says, and the jsonb index removal is the half that actually unblocks a stuck indexer: a btree over a jsonb column rejects any row whose index tuple gets too big (`index row size 4496 exceeds btree version 4 maximum 2704`). The value itself is perfectly legal, so no amount of writer-side validation can save it. It trips on incompressible content rather than on big JSON, which is why it looks intermittent in production. If that's the "writes stopping our indexers", this is it.

One thing I'd fix before merge (`value !== null` in codegen) plus a deploy note, inline.

Nothing here touches ids, dedup or conflict handling, so it won't move the `delegatedVotes` drift.

Approving.

tony8713:
Fixed: a number into a Decimal field, a number into an Int-keyed ref and `undefined` into a nullable String all persist as they do on master again, and the 256-char / NUL / 256-emoji behaviour is unchanged.

Also correcting myself on the test thread: the jsonb change is already covered, reverting that index line fails the `createEntityStores` snapshot. That half of my ask was wrong.

